PR #4, opened 2020-05-16
Major change to support subject areas

jeremiahhansen:
I recently had a customer reach out who ran into an issue with snowchange, managing cross-database dependencies. He was trying to create a view in one database which was dependent upon a table in another database (which is pretty common). Previously every database was independent of each other. I have also had people ask if it were possible for multiple teams to work independently when they didn't have dependencies on each other.

So I came up with this concept of a subject area. Essentially now a change history table is created per Snowflake account (not per database) and there is an attribute for subject area. See the updated README.md for more details.


jamesweakley:
Just so I understand correctly, is the concept of subject area being addressed to reduce versioning confusion?
In other words, it would work fine without it, but it there would be a greater risk of multiple different branches all working on what they think will be, say, v1.8.1?
And if so, would the recommended convention be that you generally limit yourselves to a single branch per subject area?

jeremiahhansen:
No, sorry if I didn't explain it well enough. The primary purpose for this change is to support cross database dependencies (like a view in one database referencing a table in another). This came up from a Snowflake customer trying to do exactly that. The way snowchange worked it kept a single metadata table per database and only managed dependencies between objects in a single database. So the way to allow cross database dependencies, as I see it, is to have the metadata table outside the databases in its own place (hence the "METADATA" database now).

The best term I could come up with for this cross-database concept was "subject area". I thought about other terms like "project" but that is temporary and more applicable to a branch with is temporary.

I debated about removing the convention of having the database name be the second level folders and am still going back and forth. Not sure how best to support having multiple accounts in the same Snowflake account because it requires that database names have a suffix or prefix. The more I think about it now though, I don't think what I have solves the problem. Maybe it's better to remove that second level requirement and make use of a placeholder/macro concept (which most DCM tools support). Could also just depend on dbt for that.

Maybe we should setup a quick Zoom call to discuss?

README.md
```diff
@@ -79,24 +74,29 @@ For example, a script name that follows this convention is: `V1.1.1__first_chang
 
 Every script within a database folder must have a unique version number. snowchange will check for duplicate version numbers and throw an error if it finds any. This helps to ensure that developers who are working in parallel don't accidently (re-)use the same version number.
 
-### Change History Table
+### Script Requirements
+
+snowchange is designed to be very lightweight and not impose to many limitations. Each change script can have any number of SQL statements within it and must supply the necessary context, like database and schema names. The context can be supplied by using an explicit `USE <DATABASE>` command or by naming all objects with a three-part name (`<database name>.<schema name>.<object name>`). snowchange will simply run the contents of each script against the target Snowflake account, in the correct order.
```
jamesweakley, line 79: I might be missing something, but is this potentially confusing when combined with the `environment-name` parameter?

It looks like that parameter is only going to affect the naming of the metadata database, but until we get templating, scripts will always apply to the fixed database & schema regardless of environment setting.

jeremiahhansen: Yeah, I did think about that. Basically I changed the meaning of the `append-environment-name` parameter to append the environment name to the metadata database not the user created databases. Even with user templating/macros snowchange will always need to manage the change history table itself, and we'll need a way for users to specify which database/schema/table they want to store the changes in.

The other way to approach this, which might be more flexible would be to remove the `environment-name` and `append-environment-name` parameters entirely and allow the user to specify/override the change history table location. It looks like flyway supports this with a configuration setting where the user can explicitly define the table name. We'd probably need to allow setting/overriding any of the database/schema/table names.

Which option do you like better? Option #2 would make snowchange even simpler which is always nice (as we're not using the environment name for anything else, like you noticed).

jamesweakley: Yeah I like option 2, would be much clearer to a new user

PR #11, opened 2020-08-20
Add variable support to snowchange

jeremiahhansen:
The goal of this feature is to allow for variable support in change scripts. This is critical to enabling people to use snowchange to support multiple environments (dev, test, prod) in the same Snowflake account. In that case the database names (and other account level object names) must include an environment name either as a prefix or suffix. This is a very common use case and is currently not possible to support with snowchange. Most database change management (DCM) tools provide functionality like this.

I think it would be very cool if we completely incorporated Jinja into snowchange but I'm not sure exactly how to do that. So for this first iteration I've followed the Jinja syntax for the placeholders but simply use regular expressions to do the replacements. Down the road it should be very easy to replace that logic with Jinja if needed.

Please see the README.MD changes for an overview of how everything is supposed to work.

### DIFF
--- a/snowchange.py
+++ b/snowchange.py
@@ -1,4 +1,5 @@
 import os
+import string
 import re
 import argparse
 import json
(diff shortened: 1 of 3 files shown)


snowchange.py
```diff
@@ -222,6 +227,32 @@ def apply_change_script(script, change_history_table, autocommit, verbose):
...
+  # Loop through each variable and replace it
+  for key in vars:
+    content = re.sub(r"\{\{ " + key + r" \}\}", vars[key], content)
+
+  # If there are any leftover variables through an error now
+  if count_variable_references(content) > 0:
+    raise ValueError("Found variables remaining in script content")
+
+  return content
+
```
jamesweakley, line 255: I think it'd probably be better to use a [safe_substitute ](https://docs.python.org/3/library/string.html#string.Template.safe_substitute) approach instead of a manual regexp solution.

There's a couple of examples [here](https://stackoverflow.com/questions/34360603/python-template-safe-substitution-with-the-custom-double-braces-format) where string.Template has the delimiter customised.

jamesweakley (changes requested):
Aside from the one comment, looks great and thanks for submitting.

snowchange.py
jeremiahhansen: Alright, updated the substitution approach to use the string Template approach. Although I used substitute instead because I want the script to fail if there are left over variables in the script (since they'll for sure fail in Snowflake).

Also, I removed the region parameter as it's been deprecated in the Snowflake Python driver (https://docs.snowflake.com/en/user-guide/python-connector-api.html#connect).

jamesweakley: Nice! looks great

jamesweakley (approved):

PR #12, opened 2020-10-05
added support for repeatable change scripts such as stored procedures…

kedar-waval:
snowchange is great at handling versioned script changes useful in maintaining objects that get altered over a period of time.
I have modified the utility a little, to make it easy to maintain objects like stored procedures and functions that could be called as "code" and always need to be applied in their entirety. Scripts that hold such code are called out as repeatable.


jeremiahhansen:
Thank you for thinking about this @kedar-waval. I have also thought about adding support for repeatable migrations to snowchange so happy to see someone take this up. But I would like us to follow the Flyway semantics for repeatable migrations. The file naming semantics for snowchange have been modeled after Flyway and I would like to keep that. Please see the documentation here: https://flywaydb.org/documentation/concepts/migrations#repeatable-migrations

This will also make snowchange better handle files in the project root folder. I think we should ignore any files that don't match the specific project files we're looking for, instead of treating all .sql files that don't start with V as repeatable.

Hope that makes sense.

kedar-waval:
Thanks for reviewing @jeremiahhansen!
Flyway style naming convention makes sense for repeatable scripts. I have made changes so that snowchange now uses that naming convention.

kedar-waval:
Hi @jeremiahhansen,
I had noticed this comment in the script:
# TODO: Figure out how to directly SELECT the max value from Snowflake with a SQL version of the sorted_alphanumeric() logic

I have modified the SELECT statement in fetch_change_history to take advantage of the INSTALLED_ON column for getting the last applied version number.

kumarjoshi:
Thank you @kedar-waval. Do you have some examples of repeatable scripts?

kedar-waval:
@kumarjoshi, did you mean examples of naming repeatable scripts?
Here you go:
R__sp_add_sales.sql
R__fn_get_timezone.sql
R__fn_sort_ascii.sql

kumarjoshi:
> @kumarjoshi, did you mean examples of repeatable scripts names?
> Here you go:
> R__sp_add_sales.sql
> R__fn_get_timezone.sql
> R__fn_sort_ascii.sql


I meant the actual scripts (similar to the ones in demo/citibike)

snowchange.py
```diff
@@ -205,7 +217,7 @@ def create_change_history_table_if_missing(change_history_table, autocommit, ver
   execute_snowflake_query(change_history_table['database_name'], query, autocommit, verbose)
 
 def fetch_change_history(change_history_table, autocommit, verbose):
-  query = 'SELECT VERSION FROM {0}.{1}'.format(change_history_table['schema_name'], change_history_table['table_name'])
+  query = "SELECT VERSION FROM {0}.{1} where SCRIPT_TYPE = 'V' order by INSTALLED_ON desc limit 1".format(change_history_table['schema_name'], change_history_table['table_name'])
```
kumarjoshi, line 220: This is not a big deal as the change_history_table won't grow in the magnitude of gbs, but something I have noticed in huge tables in Snowflake. 

```SELECT "VERSION" FROM SNOWCHANGE.CHANGE_HISTORY WHERE INSTALLED_ON = (SELECT MAX (INSTALLED_ON) FROM SNOWCHANGE.CHANGE_HISTORY) ;
  ```
type of query has better performance than

```
SELECT "VERSION" FROM  SNOWCHANGE.CHANGE_HISTORY ORDER BY INSTALLED_ON DESC LIMIT 1;
```
May be because Snowflake caches MAX and MIN statistics of the columns ?

**Comment by jeremiahhansen** [comment]
Thanks @kedar-waval and @kumarjoshi, appreciate the help here. I'll take a closer look at the changes.

With respect to the fetch_change_history() changes, I'm trying to think through the edge cases with relying on the INSTALLED_ON data instead of the version number. I want to be sure that the most recent record based on INSTALLED_ON will also return the correct version number. One thing that came to mind is that most DCM tools include the ability to undo or roll-back transactions. If snowchange supports this someday then we wouldn't always want the most recent versioned change record. Maybe at that point we'd just include the versioned and undo scripts in the query. I'll give it some more thought.

File: snowchange.py [hunk 2]
```diff
@@ -113,28 +114,39 @@ def get_all_scripts_recursively(root_directory, verbose):
...
       # Add this script to our dictionary (as nested dictionary)
       script = dict()
       script['script_name'] = file_name
       script['script_full_path'] = file_full_path
-      script['script_type'] = script_name_parts.group(1)
-      script['script_version'] = script_name_parts.group(2)
-      script['script_description'] = script_name_parts.group(3).replace('_', ' ').capitalize()
+      script['script_type'] = script_type
+      script['script_version'] = None if script_type == 'R' else script_name_parts.group(2)
+      script['script_description'] = (repeatable_script_name_parts.group(2) if script_type == 'R' else script_name_parts.group(3)).replace('_', ' ').capitalize()
```
sfc-gh-jhansen, line 142: Please apply the same script name clean-up that we use for the versioned migrations. It makes the description more readable in the logs.

sfc-gh-jhansen (commented):
Alright, still trying to think of any edge cases with using the INSTALLED_ON date but can't think of any. Looks like just this last item. Thanks again!

snowchange.py
kedar-waval: Hi @jeremiahhansen,
Thanks for reviewing the changes! 

The same script name clean-up logic (i.e. remove underscores and capitalize) has been applied to both versioned and non versioned scripts to obtain the script_description. I am referring to line 142.

Are you referring to anything else? Am I missing anything?

Thanks!
Kedar

sfc-gh-jhansen:
Sorry for the slow response here @kedar-waval, should have time to do a final review/test in the next few days. Thanks!

sfc-gh-jhansen:
Alright, I think these changes look good. Still cautious about the change to find the max version but I think that it makes sense. And at the very least I think it practically behaves the same as it does now. Thanks again for the changes!

kedar-waval:
Sure!
No problem @sfc-gh-jhansen.
 I am glad that snowchange will now support repeatable scripts.

Cheers!
Kedar

PR #23, opened 2020-12-14
Adding Support for Key Pair Authentication

rymancl:
We have a strict requirement for using private key authentication when running Snowflake scripts in our automation pipelines. This PR for **version 2.5.0** is adding support for encrypted key pair authentication based on the [docs for the Python connector](https://docs.snowflake.com/en/user-guide/python-connector-example.html#using-key-pair-authentication).

Note that I have prioritized password authentication due to the fact it was in the script first. If this logic should be changed, let me know.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,11 @@ All notable changes to this project will be documented in this file.
 
 *The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).*
 
+## [2.5.0] - 2020-12-14
+
+### Added
+- Support for encrypted key pair authentication
+
 ## [2.4.0] - 2020-11-24
 
 ### Added
--- a/README.md
+++ b/README.md
@@ -170,7 +170,7 @@ In the event both authentication criteria are provided, snowchange will prioriti
 The Snowflake user password for `SNOWFLAKE_USER` is required to be set in the environment variable `SNOWSQL_PWD` prior to calling the script. snowchange will fail if the `SNOWSQL_PWD` environment variable is not set.
 
 #### Private Key Authentication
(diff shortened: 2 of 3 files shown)


sfc-gh-jhansen:
This is great @rymancl, and will help a lot of people. Thank you for taking the time to give back. One quick question, what do you think about renaming the new environment variable names to match the existing pattern? With the (legacy) exception of the password using "SNOWSQL_" all of the environment variables begin with "SNOWFLAKE_". Would be nice to be consistent with these new ones, unless I'm missing something. Thanks again!

rymancl:
@sfc-gh-jhansen Heh, it's funny you mention that. I actually had renamed it like you mentioned, but changed it back because I didn't want to introduce breaking changes. I'd love to get rid of the "SNOWSQL_" prefix, but I'll defer to your judgement on it.

Just let me know either way, I'm happy to change it if you'd like. 

Thanks!

sfc-gh-jhansen:
Yeah, the SNOWSQL_PWD name is legacy, from the first version of snowchange. I believe it was simply to fall inline with our SnowSQL CLI tool which uses that environment variable name. I'm inclined to leave it for now, or at least allow for both (SNOWSQL_PWD and SNOWFLAKE_PASSWORD) and mark the old one as deprecated in the docs for now. I'm good with either one, just don't want to remove it without giving folks the heads up.

rymancl:
@sfc-gh-jhansen I liked your idea of having support for both, so I've implemented and tested that. I added a deprecation notice for `SNOWSQL_PWD` in both the code and the README. It'll print a message similar to this when that environment variable is set:

>snowchange.py:165: DeprecationWarning: The SNOWSQL_PWD environment variable is deprecated and will be removed in a later version of snowchange. Please use SNOWFLAKE_PASSWORD instead.
  warnings.warn("The SNOWSQL_PWD environment variable is deprecated and will be removed in a later version of snowchange. Please use SNOWFLAKE_PASSWORD instead.", DeprecationWarning)

Let me know what you think.

Thanks!

sfc-gh-jhansen:
Looking better @rymancl, thanks so much! But I noticed that the new environment variables still aren't following the "SNOWFLAKE_" naming convention, which was my original concern/suggestion. Looks like they're still named "PRIVATE_KEY_PATH" and "PRIVATE_KEY_PASSPHRASE". Can you please update those so that all environment variables follow the same naming conventions? Thanks again.

rymancl:
@sfc-gh-jhansen Fixed!

sfc-gh-jhansen:
Awesome, thanks @rymancl, just merged.

PR #37, opened 2021-04-14
Bumping dependencies

### LINKED ISSUE #36: Please bump snowflake-connector-python to 2.4.X
This should also mean we can run `schemachange` on Python3.9 (as it's now supported within snowflake-connector-python 2.4.X)!

Could we also bump the Dockerfile to 3.9 as well if the above is correct? :)

michael-robbins:
* snowflake-connector-python to 2.4.2 and allow pip to pick the latest patch release
* bumping python to 3.9
* removed 'dont use 3.9 warning'
* added 3.8 and 3.9 trove classifiers (not sure if you'd like them all, or just 'latest')
* added in a 'Dockerfile-src' that will build schemachange from src instead of just a 'pip install schemachange'

Fixes #36

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,13 @@ All notable changes to this project will be documented in this file.
 
 *The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).*
 
+## [2.9.1] - 2021-04-15
+
+### Changed
+- Bumped snowflake-connector-python to 2.4.2 and relaxed the pip dependency to ~= (pick the latest patch release 2.4.X on install)
+- Bumped the docker container python version to 3.9 as snowflake-connector-python now supports this as of 2.4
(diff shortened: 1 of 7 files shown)


sfc-gh-jhansen:
Thanks @michael-robbins, appreciate the PR. I think these changes look good, and are certainly straight-forward. I just need to remember what it was about Python 3.9 that wasn't working :). Also, can you please change the schemachange version # to 2.9.1 and add a corresponding entry to the CHANGELOG? In the meantime, I'll test it out and try and remember what the Python 3.9 issue was.

michael-robbins:
OK I've had a go at doing the version bump and changelog entry!

I've dated the changelog entry as today, but feel free to commit in the final date if it pushes out to tomorrow or beyond!

RE: Python 3.9, I think it was just the snowflake-connector-python dependency that in v2.3 didn't support python 3.9

But there very well might be schemachange specific reasons why Python 3.9 wasn't supported before!

sfc-gh-jhansen:
Just merged validated the changes @michael-robbins and they look good (and I think you were right about the Snowflake connector dependency). Thanks for your help!

PR #41, opened 2021-05-03
Support for schema starting with a number

marcogatto1:
Added double-quotes to cli.py script to make schemachange running against schemas whose name starts with a number.


sfc-gh-jhansen:
Hi there @marcogatto1, thanks for thinking about giving back to schemachange! Unfortunately I don't feel comfortable with accepting this change and having this be the default value for schemachange. [Quoted Identifiers](https://docs.snowflake.com/en/sql-reference/identifiers-syntax.html) in Snowflake are generally not a best practice for the reasons given in our documentation, so I don't want that to be the default behaviour for the schema and table name of the change history table.

But please feel free to continue to use the change in your own environment and thanks again for contributing.

PR #54, opened 2021-07-03
Json file load

MACKAT05:
Two additional options  Let some let the Vars definition and the change history table definition  reside inside the repository.
This allows an organization to make a template  CI/CD pipeline or template  release pipeline that does not have to be configured or reconfigured when a new project is spun up as the project definition of environments and change history is pointed to a standard location within the repository and pull by the agent on execution. This also avoids any reconfiguration that would be needed on a change in scope of the project where a new set of schemas and database  might be used to represent environments or other fungible entities.


MACKAT05:
closing pull to add some typo fixes in readme

PR #55, opened 2021-07-03
Json file load

MACKAT05:
Adding options that allow existing functionality to be directed to a file location instead of a command call, Permitting zeo config pipeline templates in an organization. (with out having to be an absolute wizard with shell scripting)


sfc-gh-jhansen:
Hi there @MACKAT05, thanks for taking the time to make schemachange better, I appreciate it.

I like the idea of allowing for the variables to be defined in a file, in addition to the command line. And the code changes for that looks pretty straight-forward.

But I'm not sure about having the change history table defined there as well (or in a file at all). Are you already using/testing these changes in your environment? Either way, can you please provide a sample JSON file that shows how you intend this particular feature to be used?

Thanks again!

MACKAT05:
Sure thing,
I have not yet implemented this in our "production" deployment tools
environment as I am not clever enough yet to figure out how to pip install
from the github fork to allow it to be called that way.
I have run positive and negative tests to verify that appropriate error
messages are generated by the code and it works as expected when called
correctly from my local box with a hello world sample sql, along with
regression testing ensuring the prior functionality was not degraded on the
items I touched. It does not look like I tested the two features
together, I only ran the history table option with the inline vars option.

Regarding your second question:
I was reading over my edits to the readme and I think you're spot on with
what was bothering me when I was describing what the option was.( more
context as to what is possible should be added)

Let me also include the story for context and motivating issues.
We have been working with snowflake for about a year and a half  now with
various departments contributing and sharing in one account with a couple
of child accounts when data required more strict segregation.

We have developed several public facing data presentation applications, and
a few transactional ones.
 At first we were comfortable with the wild west style of management with a
few lone heroes architecting data during the first 6 months to a year.
However the deployments have gotten much bigger and we have had several
near misses over simple things that could have been mistake proofed.
Change control is present but it is not strictly enforced, except on
critical data deployments.
And in those cases it's done via meeting and  command execution and manual
testing of services/ data after manual deployment.


We are trying to get a general change control practice off the ground, and
automated,  but there are several obstacles,
 among them being
* unfamiliarity with tools.
* right sizing a repository, since there are several domains within one
snowflake instance .
* complications with poor naming practices that cannot not be
accommodated to allow future correction via a change control solution.
* issues/limitations (features) with the deployment tool as designed
generating frustration with new users
** also complicating this being different orgs and subcontractors
participating who easily have the option of refusing to use the tool
perfeing manual execution during meetings rather than any automation if
their perception is that there is no advantage due to drawbacks.

What I aimed to do with the structural change to how this works was to
effectively eliminate one of the needed tool proficiencies and open up the
acceptance of a change control solution by making it more approachable to
more people.
To get an idea of why someone may opt not to do this we can look at one
project under change control, here are some screenshots from  the
supporting Azure deployment pipeline.
[image: image.png]
[image: image.png]
[image: image.png]

I
As you can see, the VARS item is present and driven by multiple variables
contributing but any changes that add vars are awful to execute since the
azure tool does not provide multi-line accommodations for data like the
json text.

if i were to use the new options these would be the files i would use and
store their names / relative paths in the pipeline definition.( files also
attached)
[image: image.png]


But I think this opens up more possibilities as you could deploy branches
with the samely named json file from the same CI/CD pipeline if you were
inclined.

It also makes segmentation of a project inside a repo easier since you can
define different variable sets for different swaths of directories called
using the path option.
[image: image.png]


The use I had was copying down Production data to the Test data using the
scripts in one folder before any changes were applied by the scripts on the
other folder.
In my case I did not need to use variables and simply hard coded it. but i
could definitely see the use of a source and target variables which would
change for branched deployments that are doing parallel dev/test/prod
activity to avoid development collision when working in the same instance.

(if it's not obvious there is a strong desire to keep  the data
interaccessible on the same instance so that comparison testing and
comparison queries could be done among the environments.)

-Tom

sfc-gh-jhansen:
Thanks for the details @MACKAT05, I'll read through them. But before I do, wanted to let you know that I can't see the images (nor do I see any attachments).

MACKAT05:
attaching thru ui in github
image  
image
image
image

image

zipped json since the dialog box complained
[json.zip](https://github.com/Snowflake-Labs/schemachange/files/6759144/json.zip)

MACKAT05:
Corrected comment translated from email reply. images and attachments added in reply.

MACKAT05:
oops  wrong button pressed

sfc-gh-jhansen:
Thanks again for the details @MACKAT05 (and screenshots). I definitely understand and see the benefit of enabling schemachange to get the variables from a file. But what I still don't understand, nor see the benefit of is storing the change history table as JSON files in source control.

For the time being, what do you think about separating the two features and focusing first on the variables being defined and referenced from a file?

MACKAT05:
I'm open to it, but again the features collectively are attempting to
enable a single point of configuration. We are currently having to define a
history table per deploy step per project.  The aim here was to move that
configuration inside of the repository where the devs have competency and
dominion.  if we just did the vars each project would have to customize
their workflow with their own history table definition.


Thinking about it a little if that makes sense, the ROOT_FOLDER could also
be put in the json in a similar way.
Would a complete json of  the command line inputs make sense?
{
"ROOT_FOLDER":"./somepath",
"CHANGE_HISTORY_TABLE":"SOME.TABLE.NAME",
"VARS":{"SCHEMA":"SOME_SCHEMA", "DATABASE":"SOME_DATABASE"}
}

That Way the Keys are not as wishy-washy as the current implementation
where you have to specify the name of the key to retrieve it from the
file...

sfc-gh-jhansen:
Hey @MACKAT05, thanks for the updated PR but unfortunately this isn't quite ready to be merged. Here's a short list of things that need to be updated/corrected first:

- Extra parameters in docs (history-table-in-vars)
- Argument names don't match naming conventions (words separated with dash)
- Extra files in commit (azure-pipelines.yml)
- Code in main file doesn't follow naming conventions (all lowercase, words separated with underscores)
- Not all command line variables are overridden by JSON file (only 3 of them). We probably need to think about how best to implement this.

MACKAT05:
understood.  I will work on this this weekend and will make those
corrections.
azure-pipelines.yml is where I got clever finally and  locally published a
version with the current functionality.
regarding the keys  I had thought of adding checks for keys in the VARS
file that match the input switches but i just stuck with the variable names
as listed in the current help
however i'm hesitant to implement SNOWFLAKE_ACCOUNT or   SNOWFLAKE_USER
 as those should be attached to secrets stored in azure vault or the
equivalent which should be paired with the account
the other parameters SNOWFLAKE_ROLE, SNOWFLAKE_WAREHOUSE, SNOWFLAKE_DATABASE
in practice seems to be always be overridden by the script.
I'll add them as commented code for now along and add a block of data for
the boolean flags  that would tie out to the parameter names in main.


I have another refactor that i'm thinking about applying that would move
the authentication code into a function that gets called once and the
connection would be passed as a persistent parameter.  in turn the another
function that would execute 'alter session set query tag = '<value>'
instead of setting it on connection as it stands now.  this saves the user
who has MFA enabled from continuously swiping on their phone and hazarding
a disabled account. when testing scripts locally

sfc-gh-jhansen:
Thanks for these updates @MACKAT05, it's looking better. But I still don't see the changes I mentioned about figuring out a way to allow the other command line variables (with the exception of the password) to be overridden by JSON file (I still only see 3 of them being overridden).

Also, it looks like a few of the print() statements have been removed (like displaying the version number which is important). Please add those back.

MACKAT05:
I refactored the logging feedback from the just prior version so the replacement  of values is done then reported as a batch (87,88) instead of customizing each statement for each variable.  
This definitely has a bit of a code smell to it... less than the prior.  I suppose there is a way to create a json of the expected inputs and running some kind of clever intersection command to get the inputs and then repiping it into the schemachange(**args = Somedictionary) in main.

 the version number feedback was never removed (still present at line 46), 
i'm also not certain about adding the input string used for the  change history table as there is seperate handling/validation for that. it could be useful though for detecting issues.

for the boolean flags.  
--help
--create-change-history-table
--autocommit
--verbose
--dry-run

I did not attempt to implement these. 
1 for reason of syntactical ambiguity.... does the absence from the input json indicate an override to be false? the Other being more practical is that they don't have public names to check as keys.
and the third being that these all act and behave as  mode changers,  I have the general feeling that if those are being tweaked its more for initial code pipeline development and not a production change like adding a new variable for a new schema, not that those are invalid its simply that the accessibility to those should probably not be added from within a repository, where someone could accidently add --dry-run to the json and effectively nullify output.

I'm  a little ambivalent about enabling the account name and user as those also should be handled with some level of respect given the adjacency to security.
Let me know what your thoughts are on that and we can add those using the flag callouts as names as needed or leave them out.

sfc-gh-jhansen:
Thanks for the updates @MACKAT05. And sorry for the delay. We had two big PRs that were tangled up, but those got released last night!

I hate to do this again, but looking at this again now, I think we need to take a step back and look at what we're trying to do. While this started as a way to pass the variables in a file, it really has evolved to being able to define all schemachange settings. As such it really is a config file for schemachange. So I'd like to think about (and help) figure out how to best do this.

For example, should this be implemented in YAML? And I think we should allow all parameter settings to be specified here.

Like I said, let me look around and give it some thought. And I'm happy to help out with this.

sfc-gh-jhansen:
Alright @MACKAT05, I have some exciting news! I just published version 3.1.0 of schemachange which adds config file support. Please see PR #71 and the CHANGELOG and README for more details. I'm going to close this PR.

PR #56, opened 2021-07-19
Prevent File Separator To Be Consumed By Version

### LINKED ISSUE #26: Separator
Hello

I think there is an issue with the naming pattern taken into account by the python.
The file must be named : V1.1.1__Hello_World.sql
Thus, the version extracted will be "V1.1.1" (it is the version number before the double underscore)

But if the file is named : V1.1.1__Hello_World__My_First_Script.sql
The version will be "V1.1.1__Hello_World" instead of "V1.1.1"

Best Regards

RishabhBhatnagar:
Fixes #26
 - Regex used for version splitting from the filename was a greedy one.  It consumed the entire before the last dunder (double underscore).
 - Changing the version regex group to a non-greedy (lazy) solves the issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,13 @@ All notable changes to this project will be documented in this file.
 
 *The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).*
 
+## [2.9.4] - 2021-07-22
+
+### Changed
+- Use a lazy regex matching for splitting version tags
+- Solves the issue due to which the lib consumed extra file separators
+
+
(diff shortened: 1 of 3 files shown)


sfc-gh-jhansen:
Hey there @RishabhBhatnagar, thanks for fixing this bug!

Before merging this though you need to update the schemachange version (in both schemachange/cli.py and setup.py) and add a corresponding entry in the CHANGELOG.md file. And please pick up the new changes that were just published to schemachange first.

RishabhBhatnagar:
@sfc-gh-jhansen, the required changes has been made :)
I've made sure that the latest version has been merged before pushing.

sfc-gh-jhansen:
Since we had so many simultaneous PRs, I manually added this change to version 3.0.0.

PR #57, opened 2021-07-21
Option to enable version comparison based on full change_history

MvT-Kadenza:
New Feature:
Added a new flag (--full-compare) to enable comparison based on content of change_history table and not based on maximum versionnumber. In projects with lots of objects and multiple developers this is easier to maintain because some changes take longer to be promoted and different objects have different versions.

Bugfix:
When a file is UTF8-BOM the Byte Order Marker is removed since Snowflake doesn't like this.

### DIFF
--- a/schemachange/cli.py
+++ b/schemachange/cli.py
@@ -7,6 +7,7 @@ import time
 import hashlib
 import snowflake.connector
 import warnings
+import codecs
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives.asymmetric import rsa
 from cryptography.hazmat.primitives.asymmetric import dsa
@@ -33,7 +34,7 @@ class JinjaExpressionTemplate(string.Template):
     )
     '''
 
-def schemachange(root_folder, snowflake_account, snowflake_user, snowflake_role, snowflake_warehouse, snowflake_database, change_history_table_override, vars, create_change_history_table, autocommit, verbose, dry_run):
+def schemachange(root_folder, snowflake_account, snowflake_user, snowflake_role, snowflake_warehouse, snowflake_database, change_history_table_override, vars, create_change_history_table, autocommit, verbose, dry_run, full_compare):
   if dry_run:
     print("Running in dry-run mode")
 
@@ -96,6 +97,11 @@ def schemachange(root_folder, snowflake_account, snowflake_user, snowflake_role,
     max_published_version_display = 'None'
   print("Max applied change script version: %s" % max_published_version_display)
 
+  change_history_full = None
+  if full_compare:
+    change_history_full = fetch_change_history_full(change_history_table, snowflake_session_parameters, autocommit, verbose)
+
+
   # Find all scripts in the root folder (recursively) and sort them correctly
   all_scripts = get_all_scripts_recursively(root_folder, verbose)
   all_script_names = list(all_scripts.keys())
@@ -107,13 +113,22 @@ def schemachange(root_folder, snowflake_account, snowflake_user, snowflake_role,
   for script_name in all_script_names_sorted:
     script = all_scripts[script_name]
 
-    # Apply a versioned-change script only if the version is newer than the most recent change in the database
+    # Apply a versioned-change script only if the version is newer than the most recent change in the database and full-compare is not set
     # Apply any other scripts, i.e. repeatable scripts, irrespective of the most recent change in the database
-    if script_name[0] == 'V' and get_alphanum_key(script['script_version']) <= get_alphanum_key(max_published_version):
-      if verbose:
-        print("Skipping change script %s because it's older than the most recently applied change (%s)" % (script['script_name'], max_published_version))
-      scripts_skipped += 1
-      continue
+    if not full_compare:
+      if script_name[0] == 'V' and get_alphanum_key(script['script_version']) <= get_alphanum_key(max_published_version):
+        if verbose:
+          print("Skipping change script %s because it's older than the most recently applied change (%s)" % (script['script_name'], max_published_version))
+        scripts_skipped += 1
+        continue
+
+    # Apply a versioned-change script if the version does not exist in the change_table and the full-compare flage is set
+    if full_compare:
+      if script['script_version'] in change_history_full:
+          if verbose:
+            print("Skipping change script %s because it was already applied" % (script['script_name']))
+          scripts_skipped += 1
+          continue
 
     print("Applying change script %s" % script['script_name'])
     if not dry_run:
@@ -151,11 +166,11 @@ def get_all_scripts_recursively(root_directory, verbose):
       if script_name_parts is not None:
         script_type = 'V'
         if verbose:
-          print("Versioned file " + file_full_path)
+          print("Found : Versioned file " + file_full_path)
       elif repeatable_script_name_parts is not None:
         script_type = 'R'
         if verbose:
-          print("Repeatable file " + file_full_path)
+          print("Found : Repeatable file " + file_full_path)
       else:
         if verbose:
           print("Ignoring non-change file " + file_full_path)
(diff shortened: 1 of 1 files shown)


AuelJonas (approved):

AuelJonas:
Great work. We need that feature as well :)

ingolevin:
+1

sabumaru:
+1

sabumaru (approved):

AuelJonas:
Hi @jamesweakley and @jeremiahhansen, 
could please have a look at this?
This feature is quite important for us :) 
Thanks and best regards
Jonas

sfc-gh-jhansen:
Hey there @AuelJonas, I will take a look at this ASAP (tomorrow or Wed)

sfc-gh-jhansen:
Thanks for contributing to schemachange @MvT-Kadenza.

The BOM change looks fine but I'm not sure about the primary change here around the full compare. With versioned migrations the goal is to apply, in order, only versions larger than what has currently been deployed. This is same semantics that the open source Flyway project follows. From [Versioned Migrations](https://flywaydb.org/documentation/concepts/migrations.html#versioned-migrations): "Versioned migrations are applied in order exactly once." This change would allow schemachange to apply versioned scripts out of order, which is a significant departure from how it works today and one that I'm not comfortable with.

Unfortunately with Imperative Style database change management tools, dealing with version conflicts among multiple developers working in parallel can be a pain. The way that customers deal with this is through their branching strategy and then ensuring that the version number gets incremented correctly. schemachange does have a nice feature that if it detects two or more versioned scripts with the same version it will fail with a warning. This makes it easier to detect duplicated version numbers in your integration test environment.

AuelJonas:
Hi @sfc-gh-jhansen, 

I get your point. This is why I think it is good you need to activate it via an additional flag. 
From my point of view there are two main benefits: 
* At the moment it can happen that a Version is not applied on the Test environment because there is a higher version available, but then on production it is applied because there was a smaller version. 
e.g. prd = V_1__test.sql, test = V_3__test1.sql and dev = V_2__test2.sql
* Also with this it is easier to have your version script in different folders and handle version conflict. We are using the convention V_YYYY.MM.DD_TICKETNUMBER_VERSION . With this we are handling dependencies via the ticket number and don't need to, but a lot of effort in handling the conflicts in the versioning. 

What are your thoughts?

jamesweakley:
I can understand both points of view here, in fact originally this is how schemachange worked before @sfc-gh-jhansen introduced the Flyway convention (a decision I agreed with, as the benefits outweigh the negatives).

Really the challenge is about the merge challenges associated with branching, which there's no true escape from. You can only change the place where you deal with it.

If this change was to be merged, as @sfc-gh-jhansen pointed out, it would mean that scripts can run in an order that may not have been expected by the author.

My view is that at the very least, the new optional parameter needs to reflect the danger involved. 'full_compare' sounds innocuous, and doesn't imply changed logic. Perhaps 'ignore_version_order' would be more appropriate?

MvT-Kadenza:
The changes are still executed in the version order. That part is not really changed. I agree that is something that always needs to be done. The flag changes the way the list of to be executed scripts is retrieved. Not by using max version for all but just fetch all that do not have been applied before and apply those in order. 

In our use case we have 2000+ objects managed by about 10 developers. We have divided the files into logical structures per processing layer (ingestion, combination, publication etc) and source systems. All objects have their own folder. To have this controlled by one source control and cicd is almost impossible by only looking at the maximum version number of all objects combined. I can't say to our developers: make sure your number is correct by looking at all numbers. And what if their change is pulled back from uat. They need to change their number again?

Offcourse there is a risk involved with skipped versions in the --full-compare option but i don't see any other way to manage this.

AuelJonas:
Hi, 

I think full compare is a better naming. As it is doing a full comparison of the versioned scripts. 
To use that feature a user needs to actively add the flag. If the user just keeps the code as it nothing will change. 
Usually before adding a flag you make sure, that it will do what you are expecting. 

As @MvT-Kadenza mentioned the scripts are still executed in a alphanumeric order. 

BR
Jonas

jamesweakley:
> In our use case we have 2000+ objects managed by about 10 developers. We have divided the files into logical structures per processing layer (ingestion, combination, publication etc) and source systems. All objects have their own folder.

Honestly an environment with this many objects and teams is probably easier to manage with a tool like https://github.com/dbt-labs/dbt. Because it's declarative rather than imperative, you can avoid this issue of versioned incremental schema changes, and merge conflicts are exposed and resolved much more naturally by the git workflow.

I'll let @sfc-gh-jhansen respond to the latest comments.

MvT-Kadenza:
> 
> Honestly an environment with this many objects and teams is probably easier to manage with a tool like https://github.com/dbt-labs/dbt. Because it's declarative rather than imperative, you can avoid this issue of versioned incremental schema changes, and merge conflicts are exposed and resolved much more naturally by the git workflow.
> 

I kinda agree and kinda don't agree.. We migrated from a pure MS SQL Server solution (with dacpacs) that is declarative by nature and also there you can't just blindly do releases because in data platforms there is always an order in doing things. Some changes need an data update and a developer needs to write something for that and that something will even not always be reversible. 
schemachange allows us to manage all this in a pretty easy way. The only cumbersome part right now is the max versionnumber strategy that is too rigid when doing updates at the same time (on different objects). Doing a full-compare is one way to fix that. 

We do have everything stored and gated inside azure devops (git workflow branch implementation) and proper cicd pipelines so merging, approving and releasing is not an issue here. There are always human sanity checks in place :-).

sfc-gh-jhansen:
Thanks for the great discussion here. First, I want to be clear that this change does cause the scripts to be executed out of order. To prove the point, imagine that you've deployed versions 1.1 and 1.3 to production, and then you add 1.2 to the repo and deploy. If I'm reading the code correctly, in this case it will deploy version 1.2, which is out of order. The whole point of having version numbers is to make sure the scripts are deployed in order, and only once.

So the question is do we want schemachange to support deploying scripts once out of order? If in the real world people find that a (more) useful approach then maybe we should add support. What if instead of adding a flag, which causes schemachange to process the same set of scripts in two very different ways, we add a new script type (i.e. a new initial letter instead of "V")? It seems to me that would be more clear. Plus we already have repeatable "R" scripts, and there's another PR open right now where we're also discussing adding a new script type.

What do you think?

P.S. @MvT-Kadenza, one other suggestion for making it easier to support your 2000+ objects managed by about 10 developers might be to use multiple change history tables (maybe one per processing layer, or something similar). But that might not be enough for your environment.

MvT-Kadenza:
@sfc-gh-jhansen We discussed using multiple change history tables per processing layer but for now we opted not to do that. We do split the processing of the layers into multiple schemachange executes (by supplying a different root folder) but they currently share the same change_history table. 

I do agree that adding the proposed feature would cause that version 1.2 to be deployed later if 1.3 was already promoted. In our scenario that would never happen because we do keep track of this manually via our backlog and sprint boards. We know that some changes need to be in order and we try to enforce that with the backlog that the same object will never have 2 different changes.

As for using another script type instead of a global flag I can see your point and it does make sense. That will avoid any confusion about what the tool is doing.

My proposal
"V101.1" Versioned script. Default behavior
"F101.1"  Full scripts. These will be applied when there is no 'success' record in the change_history table for this version. They will be applied in versionorder when encountered and may be executed out of order between deployments.

jeremiahhansen:
Hey all, sorry for the slow response but I'm OOO this week with the family and have been unable to review this until now. And thanks again for the lively conversation.

For more context please read up on PR #60, but the more I think about it I don't really like the idea of adding yet another change script type, it's going to add confusion. And I think that the changes in PR #60 will meet the needs here. There is currently a bug in the repeatable "R" scripts where they run every time. With PR #60 they will only run when they haven't been or when the checksum has changed. So if you create these scripts as "R" scripts and don't ever update them in source control they will always run once which I believe is what you're ultimately after.

Will that satisfy the requirements?

AuelJonas:
This feature here is definitely needed. For us it is a lot of work to orchestrate that all versions executed in the correct order. 
It happens often that someone start with one version but then someone else already used that version or a newer version was already applied. This is really bothering us at the moment. With this feature here we only need to make sure that scripts are executed in the correct order if they have a dependency, which is for us much easier to manage.

sfc-gh-jhansen:
@AuelJonas, unless I'm missing something this can be accomplished without adding a new script type. By using repeatable "R" migrations (once the bugs have been fixed) you can have each migration script run only once (assuming you don't make changes to it). Would that not meet the requirement here?

MvT-Kadenza:
@sfc-gh-jhansen 
My original reason for creating this PR is similar to the question asked here (https://github.com/Snowflake-Labs/schemachange/issues/64). We use a similar structure and even added version numbers on objects since we have so much we need to organize it to don't loose the overview and maintainability:

```
| 1000__ingestion
  | 1001_source_system_1
    | 1001.1.table_1
      | V1001.1.20210821.1__Createtable.sql
      | V1001.1.20210822.1__Updatetable.sql
      | V1001.1.20210822.2__AlterDataInTable.sql
....
    | 1001.102.table_102
      | V1001.102.20210821.1__Createtable.sql
  | 1015_source_system_15
    | 1015.1.table_1
      | V1015.1.20210821.1__Createtable.sql
      | V1015.1.20210822.1__Updatetable.sql
      | V1015.1.20210822.2__AlterDataInTable.sql
    | 1015.86.table_86
      | V1015.86.20210821.1__Createtable.sql
| 2000__combination
  | 2001_combined_schema_1
    |2001.1_combined_object_1
      | V2001.1.20210821.1__Createtable.sql
      | R__Create_SP_ProcessNewDataForcombinedObject1.sql
....
    |2001.80_combined_object_80
      | V2001.80.20210821.1__Createtable.sql
      | R__Create_SP_ProcessNewDataForCombinedObject80.sql
      | R__Create_view_SelectLogicalDatasetForCombinedObject80.sql
```

Because of the logical folder structure we have chosen (one folder per object) we don't really care for the exact deployment order across objects, but we do care about ordering deployments per object. 

Using only R__ for all scripts kind of defies the purpose of using versions...

sfc-gh-jhansen:
Changes addressed in PR #60. Closing this PR

Issue #58, opened 2021-07-22
Provide application parameter for usage stats


jamesweakley:
Per the native connector guide, supplying an "application" parameter enables visibility of how widely used a Snowflake app is.

I think this is worth doing in this project, to gain a measure of the tool's popularity/adoption.

PR #59, opened 2021-07-22
Provide an application parameter for usage stats

jamesweakley:
See issue #58


sfc-gh-jhansen:
Looks good, thanks for adding this (didn't even realize it was there)!

Issue #49, opened 2021-06-14
Adding checksum checking to repeatable change scripts


BoAustin:
Hello,

Is there a plan to mimic the behavior of flyway's [repeatable migration](https://flywaydb.org/documentation/concepts/migrations.html#repeatable-migrations) functionality regarding verifying the checksum changes before re-applying the repeatable change script?

Our use case will involve having many repeatable scripts full of stored procedures and functions and when Snowpark comes out this is going to explode and it would be great if we didn't have to re-apply the code that didn't change each time we ran schemachange.

sfc-gh-jhansen:
Hey there @BoAustin, I think that's a great idea. As you pointed out schemachange doesn't currently make use of the checksums. If you're willing, please submit a PR with those changes and I'll take a look. Thanks!

AuelJonas:
Hi @BoAustin , 
I created the code for this issue :)

sfc-gh-jhansen:
Just published this change under 3.0.0

PR #60, opened 2021-07-22
add checksum check for repeatable scripts

AuelJonas:
I added a check for the repeatable scripts on the checksum. 
With this change repeatable scripts are only executed if there is a change in the code. 
This pull request solves #49

### DIFF
--- a/schemachange/cli.py
+++ b/schemachange/cli.py
@@ -90,7 +90,7 @@ def schemachange(root_folder, snowflake_account, snowflake_user, snowflake_role,
   change_history = None
   if (dry_run and change_history_metadata) or not dry_run:
     change_history = fetch_change_history(change_history_table, snowflake_session_parameters, autocommit, verbose)
-    d_scripts_checksum = fetch_d_scripts_checksum(change_history_table, snowflake_session_parameters, autocommit, verbose)
+    r_scripts_checksum = fetch_r_scripts_checksum(change_history_table, snowflake_session_parameters, autocommit, verbose)
 
   if change_history:
     max_published_version = change_history[0]
@@ -104,7 +104,6 @@ def schemachange(root_folder, snowflake_account, snowflake_user, snowflake_role,
   all_script_names = list(all_scripts.keys())
   # Sort scripts such that versioned scripts get applied first and then the repeatable ones.
   all_script_names_sorted =   sorted_alphanumeric([script for script in all_script_names if script[0] == 'V']) \
-                            + sorted_alphanumeric([script for script in all_script_names if script[0] == 'D']) \
                             + sorted_alphanumeric([script for script in all_script_names if script[0] == 'R'])  
 
   # Loop through each script in order and apply any required changes
@@ -120,16 +119,18 @@ def schemachange(root_folder, snowflake_account, snowflake_user, snowflake_role,
       continue
     
     # Apply only R scripts where the checksum changed compared to the last execution of snowchange
-    if script_name[0] == 'D':
+    if script_name[0] == 'R':
       # get checksum of current R script
       with open(script['script_full_path'],'r') as content_file:
         content = content_file.read().strip()
-        content = content[:-1] if content.endswith(';') else content
+        content = content[:-1] if content.endswith(';') else content        
+        # Replace any variables used in the script content
+        content = replace_variables_references(content, vars, verbose)
         checksum_current = hashlib.sha224(content.encode('utf-8')).hexdigest()
 
-        # check if D file was already executed
-        if script_name in list(d_scripts_checksum['script_name']):
-          checksum_last = list(d_scripts_checksum.loc[d_scripts_checksum['script_name'] == script_name, 'checksum'])[0]
+        # check if R file was already executed
+        if script_name in list(r_scripts_checksum['script_name']):
+          checksum_last = list(r_scripts_checksum.loc[r_scripts_checksum['script_name'] == script_name, 'checksum'])[0]
         else:
           checksum_last = ''
 
@@ -171,7 +172,6 @@ def get_all_scripts_recursively(root_directory, verbose):
       file_full_path = os.path.join(directory_path, file_name)
       script_name_parts = re.search(r'^([V])(.+)__(.+)\.(?:sql|SQL)$', file_name.strip())
       repeatable_script_name_parts = re.search(r'^([R])__(.+)\.(?:sql|SQL)$', file_name.strip())
-      delta_script_name_parts = re.search(r'^([D])__(.+)\.(?:sql|SQL)$', file_name.strip())
 
       # Set script type depending on whether it matches the versioned file naming format
       if script_name_parts is not None:
@@ -182,10 +182,6 @@ def get_all_scripts_recursively(root_directory, verbose):
         script_type = 'R'
         if verbose:
           print("Repeatable file " + file_full_path)
-      elif delta_script_name_parts is not None:
-        script_type = 'D'
-        if verbose:
-          print("Delta file " + file_full_path)
       else:
         if verbose:
           print("Ignoring non-change file " + file_full_path)
@@ -196,11 +192,9 @@ def get_all_scripts_recursively(root_directory, verbose):
       script['script_name'] = file_name
       script['script_full_path'] = file_full_path
       script['script_type'] = script_type
-      script['script_version'] = None if script_type in ['R', 'D'] else script_name_parts.group(2)
+      script['script_version'] = None if script_type in ['R'] else script_name_parts.group(2)
       if script_type == 'R':
         script['script_description'] = repeatable_script_name_parts.group(2).replace('_', ' ').capitalize()
-      elif script_type == 'D':
-        script['script_description'] = delta_script_name_parts.group(2).replace('_', ' ').capitalize()
       else:
         script['script_description'] = script_name_parts.group(3).replace('_', ' ').capitalize()
       
@@ -344,10 +338,10 @@ def create_change_history_table_if_missing(change_history_table, snowflake_sessi
   query = "CREATE TABLE IF NOT EXISTS {0}.{1} (VERSION VARCHAR, DESCRIPTION VARCHAR, SCRIPT VARCHAR, SCRIPT_TYPE VARCHAR, CHECKSUM VARCHAR, EXECUTION_TIME NUMBER, STATUS VARCHAR, INSTALLED_BY VARCHAR, INSTALLED_ON TIMESTAMP_LTZ)".format(change_history_table['schema_name'], change_history_table['table_name'])
   execute_snowflake_query(change_history_table['database_name'], query, snowflake_session_parameters, autocommit, verbose)
 
-def fetch_d_scripts_checksum(change_history_table, snowflake_session_parameters, autocommit, verbose):
+def fetch_r_scripts_checksum(change_history_table, snowflake_session_parameters, autocommit, verbose):
   query = f"SELECT DISTINCT SCRIPT, FIRST_VALUE(CHECKSUM) OVER (PARTITION BY SCRIPT ORDER BY INSTALLED_ON DESC) \
           FROM {change_history_table['schema_name']}.{change_history_table['table_name']} \
-          WHERE SCRIPT_TYPE = 'D' AND STATUS = 'Success'"
+          WHERE SCRIPT_TYPE = 'R' AND STATUS = 'Success'"
   results = execute_snowflake_query(change_history_table['database_name'], query, snowflake_session_parameters, autocommit, verbose)
 
   # Collect all the results into a dict
@@ -381,13 +375,13 @@ def apply_change_script(script, vars, default_database, change_history_table, sn
     content = content_file.read().strip()
     content = content[:-1] if content.endswith(';') else content
 
+  # Replace any variables used in the script content
+  content = replace_variables_references(content, vars, verbose)
+
   # Define a few other change related variables
   checksum = hashlib.sha224(content.encode('utf-8')).hexdigest()
   execution_time = 0
   status = 'Success'
-
-  # Replace any variables used in the script content
-  content = replace_variables_references(content, vars, verbose)
 
   # Execute the contents of the script
   if len(content) > 0:
--- a/schemachange/cli.py
+++ b/schemachange/cli.py
@@ -199,7 +199,7 @@ def get_all_scripts_recursively(root_directory, verbose):
       if script_type == 'R':
         script['script_description'] = repeatable_script_name_parts.group(2).replace('_', ' ').capitalize()
       elif script_type == 'D':
-        script['script_description'] = delta_script_name_parts.group(2).replace('_', ' ').capitalize()
+        script['script_description'] = delta_script_name_parts.group(3).replace('_', ' ').capitalize()
       else:
         script_name_parts.group(3).replace('_', ' ').capitalize()
       
--- a/schemachange/cli.py
+++ b/schemachange/cli.py
@@ -199,7 +199,7 @@ def get_all_scripts_recursively(root_directory, verbose):
       if script_type == 'R':
         script['script_description'] = repeatable_script_name_parts.group(2).replace('_', ' ').capitalize()
       elif script_type == 'D':
-        script['script_description'] = delta_script_name_parts.group(3).replace('_', ' ').capitalize()
+        script['script_description'] = delta_script_name_parts.group(2).replace('_', ' ').capitalize()
       else:
         script_name_parts.group(3).replace('_', ' ').capitalize()
       
--- a/schemachange/cli.py
+++ b/schemachange/cli.py
@@ -14,7 +14,7 @@ from cryptography.hazmat.primitives.asymmetric import dsa
 from cryptography.hazmat.primitives import serialization
 
 # Set a few global variables here
-_schemachange_version = '2.9.3'
+_schemachange_version = '3.0.0'
 _metadata_database_name = 'METADATA'
 _metadata_schema_name = 'SCHEMACHANGE'
 _metadata_table_name = 'CHANGE_HISTORY'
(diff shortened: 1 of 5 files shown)


sabumaru:
I would like to see this as a nother type of file-version
something like "D" for Delta or "I" for incremental
we have scripts that need to run EACH release... 
and some that only need to run once and then when changed .. but can run every time

thanx in advance

schemachange/cli.py
```diff
@@ -114,6 +116,22 @@ def schemachange(root_folder, snowflake_account, snowflake_user, snowflake_role,
...
       continue
+    
+    # Apply only R scripts where the checksum changed compared to the last execution of snowchange
+    if script_name[0] == 'R':
+      # get checksum of current R script
+      with open(script['script_full_path'],'r') as content_file:
+        content = content_file.read().strip()
+        content = content[:-1] if content.endswith(';') else content
+        checksum_current = hashlib.sha224(content.encode('utf-8')).hexdigest()
+        checksum_last = list(r_scripts_checksum.loc[r_scripts_checksum['script_name'] == script_name, 'checksum'])[0]
```
MvT-Kadenza, line 127: This fails if the R__ script runs for the first time?

AuelJonas:
> I would like to see this as a nother type of file-version
> something like "D" for Delta or "I" for incremental
> we have scripts that need to run EACH release...
> and some that only need to run once and then when changed .. but can run every time
> 
> thanx in advance

Hi, 
thanks for the feedback. I like the idea. 
I will update the code.

MvT-Kadenza:
> 
> 
> > I would like to see this as a nother type of file-version
> > something like "D" for Delta or "I" for incremental
> > we have scripts that need to run EACH release...
> > and some that only need to run once and then when changed .. but can run every time
> > thanx in advance
> 
> Hi,
> thanks for the feedback. I like the idea.
> I will update the code.

I like the idea as well. 
Suggestion:
V - Versioned
R - Repeatable (only when checksum changed)
A - Always

AuelJonas:
> > > I would like to see this as a nother type of file-version
> > > something like "D" for Delta or "I" for incremental
> > > we have scripts that need to run EACH release...
> > > and some that only need to run once and then when changed .. but can run every time
> > > thanx in advance
> > 
> > 
> > Hi,
> > thanks for the feedback. I like the idea.
> > I will update the code.
> 
> I like the idea as well.
> Suggestion:
> V - Versioned
> R - Repeatable (only when checksum changed)
> A - Always

I like the naming. Only thing from my perspective is, that you need than to migrate your R scripts to A. 
With D we have just a new feature which can be used.

AuelJonas:
I now changed from R to D. And left R as it was before.

schemachange/cli.py
AuelJonas: Thanks for the review. I fixed this with the new commit.

schemachange/cli.py
```diff
@@ -184,8 +195,14 @@ def get_all_scripts_recursively(root_directory, verbose):
...
       script['script_type'] = script_type
-      script['script_version'] = None if script_type == 'R' else script_name_parts.group(2)
-      script['script_description'] = (repeatable_script_name_parts.group(2) if script_type == 'R' else script_name_parts.group(3)).replace('_', ' ').capitalize()
+      script['script_version'] = None if script_type in ['R', 'D'] else script_name_parts.group(2)
+      if script_type == 'R':
+        script['script_description'] = repeatable_script_name_parts.group(2).replace('_', ' ').capitalize()
+      elif script_type == 'D':
+        script['script_description'] = delta_script_name_parts.group(2).replace('_', ' ').capitalize()
+      else:
+        script_name_parts.group(3).replace('_', ' ').capitalize()
```
MvT-Kadenza, line 204: should be:
`        script['script_description'] = script_name_parts.group(3).replace('_', ' ').capitalize()`

AuelJonas: updated with new commit

schemachange/cli.py
```diff
@@ -199,7 +199,7 @@ def get_all_scripts_recursively(root_directory, verbose):
       if script_type == 'R':
         script['script_description'] = repeatable_script_name_parts.group(2).replace('_', ' ').capitalize()
       elif script_type == 'D':
-        script['script_description'] = delta_script_name_parts.group(2).replace('_', ' ').capitalize()
+        script['script_description'] = delta_script_name_parts.group(3).replace('_', ' ').capitalize()
       else:
         script_name_parts.group(3).replace('_', ' ').capitalize()
```
sabumaru, line 204: i think the bug was on this line

schemachange/cli.py
```diff
@@ -199,7 +199,7 @@ def get_all_scripts_recursively(root_directory, verbose):
       if script_type == 'R':
         script['script_description'] = repeatable_script_name_parts.group(2).replace('_', ' ').capitalize()
       elif script_type == 'D':
-        script['script_description'] = delta_script_name_parts.group(2).replace('_', ' ').capitalize()
+        script['script_description'] = delta_script_name_parts.group(3).replace('_', ' ').capitalize()
```
sabumaru, line 202: the issue was not on this line from what i can see

schemachange/cli.py
AuelJonas: Actually it was correct. 
For d and r it should be group 2 and for v group 3. 
As with v group 2 is the version.

MvT-Kadenza: @sabumaru is right. What i meant was that else is wrong...
`else:
        script_name_parts.group(3).replace('_', ' ').capitalize()`

should be
`else:
        script['script_description'] = script_name_parts.group(3).replace('_', ' ').capitalize()`

AuelJonas: sry I am blind

AuelJonas:
Hi @jeremiahhansen , 
if you find time, it would be great if you can also look at this :) 
Should also be ready for merging. 
BR
Jonas

sfc-gh-jhansen:
Thanks so much for the changes @AuelJonas (and feedback @sabumaru and @MvT-Kadenza), appreciate it. A couple things here.

First, I have been deliberate about trying to follow the script naming standards and high level migration behaviors of the open source Flyway project (since I think they've done a nice job on those fronts). Please see their [Repeatable Migrations page](https://flywaydb.org/documentation/concepts/migrations.html#repeatable-migrations) for the details, but their definition is "Instead of being run just once, they are (re-)applied every time their checksum changes."

@sabumaru, can you please provide a few more details about your use case, where you need to run a script every time? Flyway doesn't currently offer such a feature (that I'm aware of) so just curious to learn more. My preference would be to not add another type of migration if possible, but definitely want to understand the use case.

Second, @AuelJonas, please make sure and pick up the new changes that were just published to schemachange, update the schemachange version (in both schemachange/cli.py and setup.py), and add a corresponding entry in the CHANGELOG.md file.

AuelJonas:
Hi @sfc-gh-jhansen, 
we also have R scripts that we want to apply every time. A use case are role assignments for the Test and Dev environment. 
After the cloning of the environment we are applying the the role assignments. They need to be applied every time when schemachange is running. 
On the other hand there is code like views or procedures, which we only want to execute when changes happen. 

I would see as an additional feature and with the compatibility of other versions in mind I would add for this a new type. 
Otherwise if we just change the way of R scripts are working this could lead to problems for some users.

MvT-Kadenza:
@sfc-gh-jhansen  At our implementation the use-case for 'always run scripts every deployment' is the same as @AuelJonas: To reapply all security and roles after deployment.

sfc-gh-jhansen:
Hmm, this is a tricky one. It feels like this particular use case, setting up role/privilege assignments for pre-prod environments, might be better accomplished with a tool like [snowflake-rbgm (Rules Based Grant Management) by James Weakley](https://github.com/jamesweakley/snowflake-rbgm) or [permifrost by the GitLab Data Team](https://gitlab.com/gitlab-data/permifrost). Personally, I prefer that approach to managing permissions. See also my blog post [A New Approach to Database Change Management with Snowflake](https://jeremiahhansen.medium.com/a-new-approach-to-database-change-management-with-snowflake-8e3f0fee281) for more details/context.

That said, if you both still feel like we need to have a script type that runs every time I do think we need to introduce a new script type (preferably 'A' for always). Unfortunately it seems like the original implementation for repeatable 'R' scripts wasn't done correctly. If the goal is to follow Flyway's conventions as much as possible, which is my goal, then I missed that during the initial code review. Flyway docs make it clear that repeatable 'R' scripts are only applied if the files change.

So if we want to move forward here I would like to update the 'R' scripts to use the checksum check and introduce a new script type for always scripts. This will give folks that were using 'R' scripts as a "run always" script to easily transition. And we can explain the change in the CHANGELOG and them.

How does that sound?

MvT-Kadenza:
Good discussions :-).
I agree. If you purely look at the function of schemachange (applying changes), an always run update-security-script has no place inside schemachange and applying security should be done in a separate step in the deployment pipeline.

Having said that, the option to have an 'always run' feature is a nice to have i you do want to do this kind of stuff in one go and i agree with @sfc-gh-jhansen about his proposal

V - Versioned
R - Repeatable (only runs when checksum changed)
A - Always

jcollum:
> an always run update-security-script has no place inside schemachange and applying security should be done in a separate step in the deployment pipeline.

I don't agree. We are looking at Permifrost and deciding if we need to add YAT (yet another tool). I always want to avoid YAT until it's clear I need it. Schemachange repeatable scripts are fine for this, that way we know the permissions are reapplied after every deploy, in case a script screwed things up. 

I'm still having trouble seeing why a script that runs when the checksum has changed *wouldn't* be a versioned script. 

One other thing: the checksum route makes it difficult for me, as a DBA, to tell which script has been applied to a db. If the scripts are R or V it's clear what happened: the V scripts were applied if they are new and the R script was applied at every migration. This weakens the use case of having that metadata table at all.

ingolevin:
> Unfortunately it seems like the original implementation for repeatable 'R' scripts wasn't done correctly. If the goal is to follow Flyway's conventions as much as possible, which is my goal, then I missed that during the initial code review. Flyway docs make it clear that repeatable 'R' scripts are only applied if the files change.

1. If the existing implementation of R-scripts is as @sfc-gh-jhansen mentions indeed different from the flyway convention, then that almost a bug, and it should be re-aligned with the flyway conventions. **HOWEVER: this will be a breaking change for all folks who are used to and rely on R-scripts executing **_every time_**. Normally one wouldn't do these kinds of changes that break backward compatibility, but since it is essentially a bug fix, I think this is permissable. Just needs to be communicated prominently in the CHANGELOG.**
2. Adding another A-script type for "always-run" is then a useful new feature to be added. Folks will have to migrate their always-run scrips to A-scripts. 

Thus:
V - Versioned  (remains as is)
R - Repeatable (changed; executes only if checksum changed)
A - Always (new script type; executes each time)



> I'm still having trouble seeing why a script that runs when the checksum has changed wouldn't be a versioned script.

@jcollum the typical use cases for the new R-scripts (run when checksum changed) are CREATE OR REPLACE views, procedures etc.

When I make changes on these, I do not want to create an _entirely new_ sql script file (`V1__create_myView.sql`, `V2__create_myView.sql`) because then a  code reviewer in Git is not able to track what part changed from V1 to V2! it's a whole new script file and all lines will be detected by git as `Added`, while in reality I might have only changed a few lines of code compared to V2.

jcollum:
> are CREATE OR REPLACE views, procedures etc... because then a code reviewer in Git is not able to track what part changed from V1 to V2

I'd say if you are altering versioned scripts you're doing it wrong. Those scripts are like layers of an onion: first you put down a layer, then you add another layer if you need to change. Otherwise why not just have one script that does all your setup? 

Really, to be strict about it, you shouldn't be doing "OR REPLACE" in versioned scripts. It should be "CREATE" in v1 then "ALTER" in v2. 

There's two possible approaches here: Repeatable scripts that you modify incrementally (and run every time) or Versioned scripts that you do not change, you simply add a new version to make the alterations you want. If you don't want to do versioned scripts at all then Flyway isn't the right tool.

ingolevin:
> The typical use cases for the new R-scripts (run when checksum changed) are CREATE OR REPLACE views, procedures etc.

I was not talking about versioned scripts (V), but R-scripts. It's these repeatable scripts for which we want a checksum check, and run it only if changed (and not every time). 

but I realize that the filename I've used to illustrate my example started with V, which was incorrect. I will edit the comment to avoid confusion

AuelJonas:
Hey thank you all for the good discussion :) 
I think this really a hard decision. 
Just to summarize the two options: 
- Option 1: Keep R scripts as they are now and add D scripts which work like R scripts but only run, if the checksum is changing. 
Pro: This guarantees backward compatibility to the current version of schemachange. 
Con: This would go away from the actual implementation of flyaway. 
- Option 2: Adjust R scripts to only run if the checksum is changed and add a new type A scripts wich are running always. 
Pro: This keeps the functionality of schemachange close to the implementation of flyaway. 
Con: There would be no backward compatibility to previous versions (e.g. for the use of applying permissions with every run)

I personally think that backward combability is the higher value here and that flyaway is "just" an orientation on this implementation, but I leave the final decision to @sfc-gh-jhansen and @jamesweakley . 

@sfc-gh-jhansen thanks for the recommendation of snowflake-rbgm. We will look into that. 

What is your final decision?

jcollum:
@AuelJonas did you mean "A" scripts? you said "D" and that's a new letter introduced. 

I like Option 1. I use repeatable scripts to check deployments. For instance, I want to be able to confirm that a read-only user can't insert into a table and check the existence of tables. Think of it like a unit-test for a database. I got this idea after reading about R scripts (where it said they run every time) and I was like "what could I use that for?"

jamesweakley:
I'm in favour of option 1, and if @jeremiahhansen doesn't object then let's get it merged.

I appreciate the discussion and glad we've reached consensus. There is definitely a core user base who want clear Flyway style behaviour, and having a new script type to distinguish different behaviour makes the most sense to me.

MvT-Kadenza:
I am in favor of option 3:

Create the new scripttype option but make the letters configurable via a config file. This keeps it line with flyway where this is configurable as well and creates room for future new types (like the one from my pr ;-) ). This gives the implementation the option to use other letters for backwards compatibility.
 
Script types:
- versioned
- repeatable (if checksum changed)
- alwaysrun

AuelJonas:
> I am in favor of option 3:
> 
> Create the new scripttype option but make the letters configurable via a config file. This keeps it line with flyway where this is configurable as well and creates room for future new types (like the one from my pr ;-) ). This gives the implementation the option to use other letters for backwards compatibility.
> 
> Script types:
> 
> * versioned
> * repeatable (if checksum changed)
> * alwaysrun

so you are in favor for Option 2 :)

AuelJonas:
> @AuelJonas did you mean "A" scripts? you said "D" and that's a new letter introduced.
> 
> I like Option 1. I use repeatable scripts to check deployments. For instance, I want to be able to confirm that a read-only user can't insert into a table and check the existence of tables. Think of it like a unit-test for a database. I got this idea after reading about R scripts (where it said they run every time) and I was like "what could I use that for?"

For Option 1 we would add D (Delta) scripts which run only if the checksum changes and leave R as they are now. 
For Option 2 we change R to run only when the checksum changes and add A script type.

jeremiahhansen:
Hey all, sorry for the slow response but I'm OOO this week with the family and have been unable to review this until now. And thanks again for the lively conversation.

Let's go with option 2 here:
V - Versioned (remains as is)
R - Repeatable (changed; executes only if checksum changed)
A - Always (new script type; executes each time)

I believe this will provide the options that everyone is asking for, just implemented differently than some folks are suggesting. The key consideration here for me is that it sticks with Flyway semantics, upon which schemachange was based (for good reason I believe as it's a mature tool with a flexible approach).

We should make it clear in the CHANGELOG that this is a breaking change and for those relying on the current (buggy) behavior they should switch to the new "A" scripts.

Thanks everyone!

ladislav-sindelka:
I would like to bring to your attention that per Flyway standards, scripts checksum must be calculated **after** substitution of placeholders.

Current implementation of Snowchange does opposite - it computes hash of the files and after that it replaces placeholders with desired values. 
This will lead to situations where users used to Flyway standards will provide the variables to repetable script, run migration, change the variable values, run migration again but nothing will happen because {{ my_placeholder }} will always result into same hash.

By following the Flyway standard, you don't need "A" scripts...by simply providing 
`-- {{ current_timestamp }}`  comment into your repetable scripts you ensure that your repetable scripts runs always, because the timestamp will change all the time (if you provide it in --vars ofc...)

Read following for more information:
https://flywaydb.org/blog/flyway-timestampsandrepeatables
_"As the placeholder contains the current time, it changes on each migration. This changes the checksum for the repeatable migrations, meaning they get executed each time you run migrate."_

sfc-gh-jhansen:
Thank you so much for the feedback @ladislav-sindelka, that's incredibly helpful.

I definitely agree that we need to move the hashing logic to after the variable replacement, great catch on that.

As for whether we add another script type "A" or use a current timestamp on kind of on the fence. I lean towards adding a new "A" type because it seems more clear (the Flyway approach seems a bit hacky). There really do seem to be three different types of script application:

- Versioned: Only ever applied once, and in increasing version order
- Repeatable: Applied initially and every time the checksum changes thereafter
- Always: Applied every time

@AuelJonas, does that all make sense?

AuelJonas:
@sfc-gh-jhansen I know added the fix to the R scripts with the adjustments suggested by @ladislav-sindelka . 
I also added the A scripts which are executed with every run. 
Please have look at the changes. 

Sry, I had a lot of stuff to do the last week and found no time to do the adjustments :) 

BR Jonas

schemachange/cli.py
```diff
@@ -14,7 +14,7 @@
 from cryptography.hazmat.primitives import serialization
 
 # Set a few global variables here
-_schemachange_version = '2.9.3'
+_schemachange_version = '3.0.0'
```
jcollum, line 17: yes definitely agree that this is a major version change

sfc-gh-jhansen:
Looks good, thank you @AuelJonas!

Issue #52, opened 2021-07-01
When using private key authentication the key is required to have a passphrase


zchesty:
Sometimes a private key may not have a passphrase for it. schemachange will not allow a private key without a passphrase to be used.

sfc-gh-jhansen:
Hi there @zchesty, thanks for raising this issue. If you have a suggest fix to the issue, please submit a PR and I'd be happy to review it.

jcollum:
If you can avoid a private key without a passphrase you definitely should.

ladislav-sindelka:
```
    with open(os.environ["SNOWFLAKE_PRIVATE_KEY_PATH"], "rb") as key:
      p_key= serialization.load_pem_private_key(
          key.read(),
          password = os.environ['SNOWFLAKE_PRIVATE_KEY_PASSPHRASE'].encode(),
          backend = default_backend()
      )
```
changed to
if len(os.getenv('SNOWFLAKE_PRIVATE_KEY_PASSPHRASE', '')) > 0:
    private_key_password = os.environ['SNOWFLAKE_PRIVATE_KEY_PASSPHRASE'].encode()
else:
    private_key_password = None
with open(os.environ["SNOWFLAKE_PRIVATE_KEY_PATH"], "rb") as key:
  p_key= serialization.load_pem_private_key(
      key.read(),
      password = os.environ['SNOWFLAKE_PRIVATE_KEY_PASSPHRASE'].encode(),
      backend = default_backend()
  )
```

by setting password to None, the serialization uses no password and does not crash.

ladislav-sindelka:
opened PR https://github.com/Snowflake-Labs/schemachange/pull/65 for this

SlightThought:
Hey all, I believe this PR broke detection of the SNOWFLAKE_PRIVATE_KEY_PATH environment variable. When I define that env var, it errors out with "raise ValueError("Unable to find connection credentials for private key or password authentication")" on line 242. However, if I define SNOWFLAKE_PASSWORD, the cli is able to detect it and use it. I tried defining both at the same time, and it defaulted to using SNOWFLAKE_PASSWORD (expected behaviour, but showing that it is not an issue with discovery of my env vars).

SlightThought:
Update regarding prior comment:

[...]

PR #65, opened 2021-08-06
add support for private keys that are not encrypted

ladislav-sindelka:
Adressing issue
https://github.com/Snowflake-Labs/schemachange/issues/52

This change adds support for private keys which are not protected(encrypted) by passphrase. 
If passphrase is not provided, it is assumed the  key is not encrypted.


sfc-gh-jhansen:
Thanks for making these changes @ladislav-sindelka. Please make sure to update the schemachange version (in both schemachange/cli.py and setup.py), and add a corresponding entry in the CHANGELOG.md file.

sfc-gh-jhansen:
Looks good to me, thanks!

Issue #72, opened 2021-09-22
missing quote around schema/database name


estheban:
Based on Snowflake documentation: https://docs.snowflake.com/en/sql-reference/sql/create-schema.html

To support schema name starting with something else than an alphabetic character (ex: starting with a number) or special character (like a `-`)

PR #73, opened 2021-09-22
Add quotes around database and schema name

estheban:
For https://github.com/Snowflake-Labs/schemachange/issues/72

### DIFF
--- a/schemachange/cli.py
+++ b/schemachange/cli.py
@@ -405,7 +405,7 @@ def get_change_history_table_details(change_history_table_override):
 
 def fetch_change_history_metadata(change_history_table, snowflake_session_parameters, autocommit, verbose):
   # This should only ever return 0 or 1 rows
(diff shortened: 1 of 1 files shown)


LeoPastaBuddy:
This PR would be super useful for me as well, as it would allow us to adopt lower case database names. 
Any plans on when this could be merged to main?

estheban:
Quick follow up, can we merge this ?

sfc-gh-jhansen:
Hi there @estheban, thanks for the reminder, and sorry for the slow response here. I'm a bit hesitant to merge these changes. IMO, it's a best practice in general to avoid quoted identifiers. See the Snowflake [Identifier Requirements](https://docs.snowflake.com/en/sql-reference/identifiers-syntax.html) page for more details (particularly the quoted identifier sections/warnings).

As a result, I'm hesitant to make this the default for all users. I'd like to better understand if/how it will impact users. Maybe we add a corresponding flag to enable this behavior and leave it disabled by default?

estheban:
Sure, a config flag is a good idea.  

Personally I moved away as for my use case, I have a lot of repetition to do and going with a coded approach was better suited for us.

PR #74, opened 2021-09-27
loosen and upgrade deps

akdor1154:
Running into issues using schemachange in a project which enforces version constraints properly.
I am using pandas 1.3 and the snowflake python connector 2.6 - this means schemachange **can't be installed** because it explicitly needs pandas <1.3 and the python connector <2.5 .

This PR relaxes the version constraints to only require matching major versions. Keeping it strict would requiring you guys to cut new versions whenever pandas or the snowflake connector publish a new minor version, which seems like needless work.

Additionally I've bumped the snowflake connector and pandas minor versions to make the intention clear, though technically there is no effect here on consumers due to the relaxed constraints.

Hope this is ok!

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
-snowflake-connector-python~=2.4.2
-pandas~=1.2.5
-pyyaml~=5.4.1
+snowflake-connector-python~=2.6
+pandas~=1.3
+pyyaml~=5.4


michael-robbins:
Would be good to get this looked at given snowflake-connector-python's now up to 2.6.2 and I'm sure others would like to upgrade to that separately as well in their code bases that need both.

Could we pin it to v2 or do Snowflake not provide any 'backwards compatibility' with their major releases of snowflake-connector-python?

sfc-gh-jhansen:
Hey there @akdor1154, so sorry for the slow response here, and thank you for your contributions! I like what you've done with this PR and PR #75. Can you please combine those into one PR and make sure you bump the schemachange version (in both the `cli.py` and the `setup.py`) and make updates to the CHANGELOG.md as well?

akdor1154:
all done, thanks!

PR #76, opened 2021-09-27
Add "--explain-first" to run `explain` on statements

akdor1154:
At our org, we like to run `explain` on DDL statements before running them. With the notable exception of GRANT, this is supported on almost all DDL statements we normally have deal with, and serves as a very capable pre-check step, e.g. for us to run in feature branch CI runs or as a pre-commit hook. (with EXPLAIN, snowflake not only checks sql syntax, but also column, table, and function references, type compatibility, and probably more)

This pr adds an `--explain-first` flag to schemachange that will run `explain <statement>` before running any statements, which enables the above workflow. This also works with dry runs.

One drawback is that not all statements support EXPLAIN - e.g. I have hit GRANT statements but there might be others. Currently this PR is an all-or-nothing approach. This is still useful for us but I could see future work enabling this on a per-statement basis. This is non-trivial though, might need to add a magic var, comment, or something else to the sql code. Would seek your design input before doing anything in this direction.

Sorry for the bombardment of PRs, this is the last one! :)

Will squash before merging, just will await feedback first in case there are other changes / you don't like this direction / etc.

Cheers
Jarrad

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,24 @@
+name: Python tests
+
+on: [push]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.7, 3.8, 3.9]
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install dependencies
+        run: |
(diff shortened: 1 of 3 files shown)


sfc-gh-tmathew:
@akdor1154 Could you revisit this PR and let us know if you could get the changes to the latest state so that we review it against the latest version and request additional enhancements for a future release. 

I can see Explain Plan being useful as part of the dry-run validations to append to a list of pre-checks.

Thoughts?

sfc-gh-twhite:
Closing this for now since the refactor and it's a bit stale.

PR #87, opened 2021-11-04
Fix tests and add CI

akdor1154:
Fixes broken main tests and adds github CI so they don't break again :)
Should be a little less fragile now, though it's still a mock based test with all the downsides they entail.

Will conflict slightly with #86  but trivial to fix on either.

CI run can be seen in my fork: https://github.com/akdor1154/schemachange/actions


akdor1154:
Closed, included with #89.

PR #88, opened 2021-11-04
move to setup.cfg

akdor1154:
This moves to setup.cfg. (https://setuptools.pypa.io/en/latest/userguide/declarative_config.html).

This is preferable over setup.py mainly because it's declarative, so tooling doesn't need to execute arbitrary code to generate metadata when trying to install this, and because it (along with included pyproject.toml) adheres to PEP-517 and PEP-518.

As an example, adding this project's git repo as a dependency to a poetry project will not install any of its dependencies, because poetry does limited AST parsing of setup.py. The recommended fix for this is to follow PEP-517/PEP-518 (in short, use pyproject.toml to tell packages how to install each other).

There should be no change in usage/dev workflow of this at all. `pip` still works exactly the same way, and package setup is still managed by setuptools. `python setup.py install` won't work anymore but no-one should be doing this anyway. I guess you are building wheels with `pip` - this too won't change.

A more violent change might be to switch to a different project manager like poetry, but I'll leave that for the repo owner :)


sfc-gh-jhansen:
Hey @akdor1154, thanks for the suggestions. There have been a few PRs merged since this one. This branch now has conflicts and you're missing the version number update (in cli.py and setup.py) and the updates in the CHANGELOG. Can you please help with those items?

And sorry, but I'm not an expert in this area. I have been running the following commands to PyPI:

```
python setup.py sdist bdist_wheel
twine upload dist/*
```

How will this change that process?

akdor1154:
Ah sure. I don't think even Guido is an expert in Python packaging, it's a bit of a minefield haha.
When migrating away from setup.py, setuptools recommend to use `python -m build` in place of `python setup.py sdist bdist_wheel` (see [link](https://packaging.python.org/tutorials/packaging-projects/#generating-distribution-archives)). This should do exactly the same thing, so you can run `twine` the same way afterwards.

akdor1154:
Bumped to 3.3.1, sound ok?

sfc-gh-jhansen:
Just merged, thanks @akdor1154!

PR #90, opened 2021-11-04
Add Jinja filter for using environmental variables in templates

tsavko:
Adding a Jinja custom filter that allows using environmental variables in templates. This is very useful for passing in secrets for external stages without logging them to STDOUT, like it is done for command line parameters.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@ All notable changes to this project will be documented in this file.
 
 *The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).*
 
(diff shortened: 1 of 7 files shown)


sfc-gh-jhansen:
Hey @tsavko, thank you so much for your contributions to schemachange! We had two different PRs with support for using environment variables in Jinja. I just published #91 so this one is redundant now. I understand that the two were slightly different approaches, so please let me know if the new features from #91 won't meet your needs for some reason. Thanks again!

tsavko:
Hey @sfc-gh-jhansen, the approach in #91 works just as well, thanks for the info!

tsavko:
Hey @sfc-gh-jhansen, #91 doesn't fully cover our needs, but I was able to easily extend it in #95. Please review that one when you get a chance. Thanks!

PR #89, opened 2021-11-04
3.2.2: fix tests, add CI, Argv as args to main() .

akdor1154:
This is a small change with a couple of benefits -

main one is you can invoke schemachange in a python script by invoking schemachange.main(['schemachange', 'deploy', ...]), which is useful.

secondary is it makes the tests a bit nicer :) I've based this on tests pr to show (and test) this, the commit to look at is the top one, aa9b503733399fd67f3f0d28437f4fe5aba752d4

Issue #84, opened 2021-11-02
Question on Jinja templating support


tsingh2k15:
Is there a support for Jinja templating inside "schemachange-config.yml" ? I think it could be useful to have the support to allow more flexible configuration as the values could vary based on environments. Thoughts?

sfc-gh-jhansen:
Yep, this has just been release with version 3.3.0 (see PR #91)

PR #91, opened 2021-11-06
3.3.0: Adding dynamic support to config file

saperry:
# Description
Extending Jinja templating support to schemachange-config.yml, plus a custom function to allow access to environmental variables. I believe this would satisfy #84.

A command line like the following
```
schemachange -f MORE2/ -a ${SF_ACCOUNT} -u ${SF_USERNAME} -r ${SF_ROLE} -w ${SF_WAREHOUSE} -d ${SF_DATABASE} -c ${SF_DATABASE}.SCHEMACHANGE.CHANGE_HISTORY --create-change-history-table 
```

could use the following schemachange-config.yml file
``` yaml
config-version: 1
root-folder: MORE2/
snowflake-account: {{env_var("SF_ACCOUNT")}}
snowflake-user: {{env_var("SF_USERNAME")}}
snowflake-role: {{env_var("SF_ROLE")}}
snowflake-warehouse: {{env_var("SF_WAREHOUSE")}}
snowflake-database: {{env_var("SF_DATABASE")}}
change-history-table: {{env_var("SF_DATABASE")}}.SCHEMACHANGE.CHANGE_HISTORY 
create-change-history-table: true
```

Since it is using the jinja template engine additional logic can also be added to vary values based on an environmental variable. The following is an example of varying the config value based on some custom logic.

``` yaml
config-version: 1
snowflake-account: {{"prod_account" if env_var("SF_ENVIRONMENT") == "PROD" else "dev_account"}}
snowflake-user: {{"PROD_USER" if env_var("SF_ENVIRONMENT") == "PROD" else "DEV_USER"}}
```

# Conflicts
This pull request does not include the ability for script files to access environmental variables, just the config file. However, I need to call out a conflict with #90, since it provides a different approach for accessing environmental variables. One exposes a custom filter and the other a custom function. We should make sure only one approach is used.

I feel the functionality provided by 'env_var' function has the edge, it allows:
1. a default value to be used if the environmental variable is not available
2. an error to be raised if the environmental variable is not available

while 'from_environ' only support the first scenario. 

If we aligned the approach used in #90 to use 'env_var' function

```
CREATE OR REPLACE STAGE TRIPS
    URL = 's3://snowflake-workshop-lab/citibike-trips'
    CREDENTIALS = (
        AWS_KEY_ID = '{{ aws_s3_access_key | from_environ("AWS_S3_ACCESS_KEY") }}',
        AWS_SECRET_KEY = '{{ aws_s3_secret_key | from_environ("AWS_S3_ACCESS_KEY") }}'
    );
````
would be rewritten as follows. 
```
CREATE OR REPLACE STAGE TRIPS
    URL = 's3://snowflake-workshop-lab/citibike-trips'
    CREDENTIALS = (
        AWS_KEY_ID = '{{ env_var("AWS_S3_ACCESS_KEY", aws_s3_access_key) }}',
        AWS_SECRET_KEY = '{{ env_var("AWS_S3_SECRET_KEY", aws_s3_secret_key) }}',
    );
```
If you wanted it to fail if the environmental variable does not exist it would be like 
```
CREATE OR REPLACE STAGE TRIPS
    URL = 's3://snowflake-workshop-lab/citibike-trips'
    CREDENTIALS = (
        AWS_KEY_ID = '{{ env_var("AWS_S3_ACCESS_KEY") }}',
        AWS_SECRET_KEY = '{{ env_var("AWS_S3_SECRET_KEY") }}',
    );

```

@sfc-gh-jhansen  let me know your thoughts on the style you would like for the jinja template to access environmental variables.

# Other considerations
Would we want to bundle this with #89 and move to 3.3.0 and one release to PYPI?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 *The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).*
 
 
+## [3.3.0] - TBD
+### Added
+- Added processing of schemachange-config.yml with jinja templating engine.
+  - Included new Jinja function env_var for accessing environmental variables from the config file.
+
 ## [3.2.1] - 2021-11-04
 ### Fixed
 - Jinja Template Engine was not recognising scripts in subfolders on windows machines. Jinja was expecting the paths to follow a unix style ie SQL/V2.0.0__ADHOC_SCRIPT.sql but on windows machines this was being passed through as SQL\V2.0.0__ADHOC_SCRIPT.sql.
--- a/README.md
+++ b/README.md
@@ -255,6 +255,22 @@ dry-run: false
 
 ```
 
+#### Yaml Jinja support
+The YAML config file supports the jinja templating language and has a custom function "env_var" to access environmental variables.
+
+##### env_var
+Provides access to environmental variables. The function can be used two different ways.
+
+Return the value of the environmental variable if it exists, otherwise return the default value.
+``` jinja
+{{ env_var('<environmental_variable>', 'default') }}
+```
+
+Return the value of the environmental variable if it exists, otherwise raise an error.
+``` jinja
+{{ env_var('<environmental_variable>') }}
+```
+
 ### Command Line Arguments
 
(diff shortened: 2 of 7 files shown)


sfc-gh-jhansen:
Yep, this looks great. I'm not a Jinja expert but I think that your approach makes sense. And it works like the [dbt env_var](https://docs.getdbt.com/reference/dbt-jinja-functions/env_var) function which I like. Thank you for the help here!

PR #93, opened 2021-11-07
Jinja2 XSS and other vulnerabilites

tsingh2k15:
- Add autoescape=True to Jinja enviornment init
- Join if statements for improving code readability


sfc-gh-jhansen:
Looks good. I'll update the CHANGELOG.

PR #95, opened 2021-11-09
Add env_var support to migration templates

tsavko:
Extending functionality provided in #91 to the migration template files.


tsavko:
Hey @sfc-gh-jhansen, the approach taken in #91 is great, however, it doesn't cover our need of accessing environmental variables directly from migration template files. This PR mimics #90 using logic added in #91.

sfc-gh-jhansen:
Looks good to me, thanks!

PR #96, opened 2021-11-09
Use default schema in connection. Option to use default DB and Schema for change history table

Pi3rral:
Added option to use a default schema when connecting to snowflake. 
Added option for change history table to use default connection parameter (database and schema).

Tests are passing with new unit tests for change history table.
It should be backward compatible.

Let me know what you think.


Pi3rral:
Is there someone to tag to get it reviewed ? Not sure how to make it visible by admins.. thanks!

Pi3rral:
@sfc-gh-jhansen would you be able to take a loot at it ? 
Thanks

sfc-gh-jhansen:
Hey there @Pi3rral, thanks for your interest in schemachange, and being willing to contribute!

Can you please explain what issue you're running up against that these changes are supposed to solve? schemachange already provides the ability to define a custom database, schema, and table name for the change history table (with the -c command line argument or in a YAML config file). Does that not work for you?

Pi3rral:
@sfc-gh-jhansen thanks for the answer, and sorry I wasn't clear on my issue.
I'm working on a multi-tenancy environment where every tenant has its own schema in the same database. So I need one change history table by tenant. That means that I need to override the name with a custom made variable every time i'm applying a migration.
This PR will help for the scalability of it by adding an option of using the default connection parameter instead of the default database and schema. Also, I'm adding the way of selecting a default schema to the connection for that.
Everything is backward compatible so it's just an option that we can use in our workflow (or not).

Pi3rral:
hey @sfc-gh-jhansen 
would you be able to review it?
thanks!

sfc-gh-jhansen:
Hi there @Pi3rral, thanks for the details. I have some concerns about adding these changes because the new flag doesn't make a lot of sense to me, especially given the -c option that already exists. If I understand these changes and your use case correctly, you're invoking schemachange for each different tenant. So you're passing in a different schema name in each invocation of schemachange, right? If so then you can accomplish your goals today with the -c option. This extra feature seems redundant to me.

But please help me see what I'm missing. Maybe give me a couple sample/cleansed examples of how you're executing schemachange (with cleansed command line arguments and values). Thanks!

Pi3rral:
hey @sfc-gh-jhansen , thanks for the comments.
You're right in the sense of it can be done with the `-c` options, but in a multi-tenancy environment, that means this argument needs to be constructed with the same info as the database connection. The new option is kind of a wrapper on top of it. It's unit tested and less prone to error.
Also this PR is about adding a schema name when creating the connection (which was needed for the table creation, but kind of independent of it).
You still think it's redundant and useless ?
Thanks.

sfc-gh-jhansen:
Hey @Pi3rral, thanks for the follow-up. Yeah, at this point I think adding this feature would be mostly redundant and potentially confusing for users.

As for the new schema parameter, that would be a fine addition if you want to remove the other code (for the default connection parameter). But I did notice that it's not used in the snowflake.connector.connect() method call, so it wouldn't do anything if you just removed the default connection parameter code.

Thanks!

Pi3rral:
sorry for the delay. I'm closing the PR as you're right, it's not using the schema parameter for connection. I'll re-open something if needed.
Thanks again for your feedback.

PR #101, opened 2021-12-07
--query-tag command line parameter

djsecrist:
add command line parameter for appending a string to the Snowflake QUERY_TAG session variable

### DIFF
--- a/schemachange/cli.py
+++ b/schemachange/cli.py
@@ -158,10 +158,12 @@ def deploy_command(config):
   print("Using default warehouse %s" % config['snowflake-warehouse'])
   print("Using default database %s" % config['snowflake-database'])
 
-  # Set default Snowflake session parameters
+  # Set default and optional Snowflake session parameters
   snowflake_session_parameters = {
     "QUERY_TAG": "schemachange %s" % _schemachange_version
   }
+  if config['query-tag']:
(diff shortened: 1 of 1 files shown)


djsecrist:
The idea here is to be able to add some traceability to every SQL statement that is executed by a CI/CD pipeline by including some additional information, such as WHO triggered the pipeline and which execution was it that caused the database change.  For example, include the UUID of the user from a Bitbucket Pipeline with $BITBUCKET_STEP_TRIGGERER_UUID.
https://support.atlassian.com/bitbucket-cloud/docs/variables-and-secrets/#Reference-variables-in-your-pipeline

sfc-gh-jhansen:
Thanks so much for contributing @djsecrist! I think the idea here is great. Just need to have a few additional files updated:

- Increment the version number in both setup.cfg file and cli.py
- Add a corresponding entry in the CHANGELOG.md file
- Update the relevant YAML config and command line sections in the README.md file with this change

Thanks again!

djsecrist:
I made the requested updates, @jeremiahhansen. Please take a look.

sfc-gh-jhansen:
Alright, looks good, thanks!

sfc-gh-jhansen:
Oops, forgot one thing @djsecrist. It looks like one or more of our tests are now failing because of this new argument. Please check out the Actions run history for more details, but it appears that at least the tests/test_main.py test needs to be updated with this new argument.

PR #102, opened 2021-12-10
Update test_main.py

djsecrist:
Added new command line parameter --query-tag


sfc-gh-jhansen:
Thanks @djsecrist! Just published 3.4.1 to PyPI.

PR #120, opened 2022-06-14
Update fetch_r_script_checksum to iterate over multiple results batches within a cursor

gcnuss:
In our experience using schemachange, we have run into an issue where fetch_r_script_checksum fails due to the quantity of R-scripts in our pipeline exceeding the data size that will fit within one Result Batch as returned by snowflake-connector-python.  It will successfully iterate over all the rows in the first "chunk" but fail to proceed into a second chunk if it exists.  The suggested code update resolves this by first returning all results batches and iterating through the rows within each batch.  It does require pyarrow to be installed, which I've added to setup.cfg.  The specified range of versions matches those required by snowflake-connector-python.

Note that I did discover the latest version of snowflake-connector-python (2.7.8) _no longer fails_ with the existing fetch_r_script_checksum code as-written, so upgrading that config requirement may also work.  That said I tested the suggested code update below with snowflake-connector-python 2.6.2, 2.7.4, and 2.7.8 and it works on all of them, so it would be more flexible and could be implemented without upgrading to 2.7.8 yet.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,6 +1,6 @@
 [metadata]
 name = schemachange
-version = 3.4.1
+version = 3.4.2
 author = jamesweakley/jeremiahhansen
 description = A Database Change Management tool for Snowflake
 long_description = file: README.md
@@ -20,6 +20,7 @@ install_requires =
     pandas~=1.3
     pyyaml~=5.4
     jinja2~=3.0
+    pyarrow>=6.0.0,<6.1.0
 include_package_data = True
 
 [options.entry_points]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@ All notable changes to this project will be documented in this file.
 
 *The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).*
 
+## [3.4.2] - 2022-06-14
+### Fixed
+- Updated fetch_r_script_checksum method in the CLI to iterate over multiple results batches within a single cursor. This protects against failure in the case where the R-script results are large enough to require multiple batches. Ref snowflake-python-connector documentation here: https://docs.snowflake.com/en/user-guide/python-connector-api.html#get_result_batches
+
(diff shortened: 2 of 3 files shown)


setup.cfg
```diff
@@ -20,6 +20,7 @@ install_requires =
     pandas~=1.3
```
gcnuss, line 20: I'm curious why the install_requires are set to "~=" instead of >= or with upper/lower bounds like I added for pyarrow.  I couldn't find other examples using "~=" online, what is the purpose of that approach?

jonrobinsdev:
+1. Am having the same issue currently with our snowflake migrations pipelines that use schemachange. Would love if this fix could be merged, or at the very least snowflake-connector-python could be updated to 2.7.8

sfc-gh-jhansen:
I'm so sorry for the delay here everyone (@gcnuss, @jonrobinsdev), I've been super busy lately and unable to give schemachange the attention it deserves. I just published a new version of schemachange (3.4.2) which updated the dependency on snowflake-connector-python to the newest version 2.8.

Please let me know if that solves the problem, and if you still think we should move forward with the batching code (and dependency on pyarrow) here.

setup.cfg
sfc-gh-tmathew: @gcnuss The pattern is aligned to [this](https://peps.python.org/pep-0440/#compatible-release) reference.

sfc-gh-tmathew:
@gcnuss  @jonrobinsdev We have been reviewing open requests and trying to cathup on open items. Would appreciate your help to address this open item.

Does the issue still exist? Does the latest version render this PR unnecessary? 

Kindly confirm

sfc-gh-tmathew:
Hi @gcnuss,

Thank you for this fix for cursor iteration with large R-script sets!

This PR has been open for over 3 years, and as you noted in the original description, upgrading `snowflake-connector-python` to 2.7.8+ resolves this issue. We've since upgraded to much newer connector versions (3.x), so this specific fix is likely no longer needed.

We're going to **close this PR** as the underlying issue has been resolved through dependency updates.

Thank you for your contribution and for identifying this issue originally!

PR #132, opened 2022-09-28
PEP8 Sorted Imports

fartzy:
Sorted the imports as in PEP8 standard.  https://peps.python.org/pep-0008/#imports

This makes the import lists a little easier to follow - and its the standard some people are used to.

I used the `isort` config but I didn't include the `.isort.cfg` because people might not want to use that one going forward.

### DIFF
--- a/schemachange/cli.py
+++ b/schemachange/cli.py
@@ -1,25 +1,25 @@
-import os
-import string
-import re
 import argparse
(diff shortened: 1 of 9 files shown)


sfc-gh-jhansen:
Taken care of in version 3.5.0, which has been published to PyPI

PR #140, opened 2023-01-29
Adding Support for SSO (Browser Based and Okta) in 3.4.3

sfc-gh-tmathew:

PR #139, opened 2023-01-28
Removed unused imports

gawalimangesh007:
Removed unused rsa and dsa imports from cryptography.hazmat.primitives.asymmetric

PR #141, opened 2023-01-29
Additional Authentication methods and Connection Refactor.

MACKAT05:
Did a number of things here in addition to adding the authentication methods.
The inversion of the authentication to a more top level part of the program removes a lot of overhead.
I do have some concerns that could save some headaches when reviewing.
1. I have been able to test the  Oauth and external browser methods fairly easily however i don't have an environment to test Okta, I included it anyway since the specification in the snowflake connector seemed straightforward to implement.

2.  I included the additional Command line element exactly like the vars command.... however i could not successfully test it... --vars using the readme example produced a similar error on windows 'Invalid json loads.'  so its not clear if the error was in what i was trying to feed it or just the windows environment.   It would be nice to include this with nested data structures handled but some kind of prefixing mechanism could be used to group data payload and header data later if it *must* be in the command line as a flat structure.  In addition the inclusion of secrets in the command line call also is a cause for concern and we may have to extricate it out.... and pull it from an environment variable later somehow....

3.  Its unclear whether the login per interaction to a singular login session would be a breaking change.   depending on what end users are using for logging besides the inbuilt mechanism.  There are 2 more transactions per script to update the query tag and instead of multiple sessions a single one....

4.  Internally I rewired  a lot of things; 
a. I moved the messages and sql to variables so it would be easier to read and audit them.  and for the ambitious users add localization. It also made it easier to  keep the width under 100 or so characters.

b. I standardized on snake_case since python cannot handle kebab-case names as arguments; doing so allowed the use of dictionary unpacking to simplify the calls.

c. I also reworked some of the logic i touched to use a more flexible pattern. now when essential checks are updated its a matter of adding an item to a list.

~~Other minor issues: i could not get the markdown links to work from my editor.~~


README.md
```diff
@@ -30,6 +30,9 @@ For the complete list of changes made to schemachange check out the [CHANGELOG](
 1. [Authentication](#authentication)
    1. [Password Authentication](#password-authentication)
    1. [Private Key Authentication](#private-key-authentication)
+   1. [Oauth Authentication] (#oauth-authentication)
```
sfc-gh-tmathew, line 33: remove the space between `] (` and the markdown reference will work.

sfc-gh-tmathew: @MACKAT05

MACKAT05: Thanks!  I don't know how those spaces got in there since I had copy pasted one of the other elements to create them.

sfc-gh-tmathew: This is some excellent work! Thank you for your contribution!

schemachange/cli.py
```diff
@@ -140,72 +186,313 @@ def redact(self, context: str) -> str:
...
+    elif os.getenv("SNOWFLAKE_AUTHENTICATOR") == 'externalbrowser' and os.getenv("SNOWFLAKE_AUTHENTICATOR"):
+      self.conArgs['authenticator'] = 'externalbrowser'
+      if self.verbose:
+        print(_log_auth_type %  'External Browser')
+        
+    elif (os.getenv("SNOWFLAKE_AUTHENTICATOR").lower()[-9:] == '.okta.com' \
+      or os.getenv("SNOWFLAKE_AUTHENTICATOR").lower()[-16:] == '.oktapreview.com'  ) \
+      and os.getenv("SNOWFLAKE_AUTHENTICATOR"):
+      okta = os.getenv("SNOWFLAKE_AUTHENTICATOR")
+      self.conArgs['authenticator'] = okta
```
sfc-gh-tmathew, line 323: @MACKAT05 Could we check for 'https://' instead of okta string. That way we can avoid limiting it two endpoint patterns but allow the end user to use the Okta endpoint set up to authenticate through SSO. I have a PR #140 for 3.4.3 enabling SSO via Okta.

sfc-gh-tmathew (commented):
@MACKAT05 Also, as part of #140 I have included requirements.txt so that the dockerfile works as designed. Updated the .gitignore file to exclude .DS_Store files. FYI.

schemachange/cli.py
MACKAT05: I think that is possible; i was not sure if okta as implemented by the connector was a generic standard. and looking into it further it does look to support that.

so `'os.getenv("SNOWFLAKE_AUTHENTICATOR").lower()[:4]=='http'`  as a single string check and include in the readme that the url must not be malformed?   
I looked for a simple python check if string is an url but I did not find one that was straightforward and did not create additional dependencies.  
would the  shorter string check to include those whom may inadvisably be testing on non https endpoints.

MACKAT05:
> 

There were a few elements from other pull requests i considered including but i was unsure of #132 and #139;  i'll go add the dockerfile items now

MACKAT05:
Looks like i am also missing a test_SnowflakeSchemachangeSession.py

schemachange/cli.py
sfc-gh-tmathew: would look for `https://` instead of just `http`. That single string check should ensure that it is secure URL. If the URL is invalid, then python connector for snowflake should return an error that the url endpoint is invalid. It is better to enforce the expectation of https than open http options.

sfc-gh-tmathew: Yes. Mentioning it in the readme file that URL should not be malformed otherwise, the authentication will not work.

MACKAT05:
> Looks like i am also missing a test_SnowflakeSchemachangeSession.py

I started to write these tests but i realized that it would be kind of hard to test the interaction without something ( as nowflake account / oauth provider)  to talk to .... how should we handle this? we could add a 'Test' mode to the Class and just have the functions return the critical outputs?

sfc-gh-jhansen:
Thanks @MACKAT05 (and @sfc-gh-tmathew), I'll take a look at this soon!

sfc-gh-jhansen:
I will review on Monday!

sfc-gh-jhansen:
Hey there @MACKAT05, I just merged these changes and noticed that the tests in the associated GitHub Action workflow failed. Can you please take a quick look to see why?

MACKAT05:
the default config needs to be updated to use snake case :/  missed that
[image: image.png]

i'll check for similar issues and push soon

MACKAT05:
@sfc-gh-jhansen  updated test Files... is a new pull request needed?

sfc-gh-jhansen:
Thanks @MACKAT05, and no need for a new pull request

sfc-gh-jhansen:
Alright @MACKAT05, I'm not run the updated version of schemachange on my laptop with my test configuration that I had been using. I dug into it a bit and think I found another bug in this release. Previously all the metadata operations/functions which called `execute_snowflake_query()` passed as the first parameter the database of the change history table. And because that method would make a new connection to Snowflake each time (bad) it also set the default database each time. And we had two different default databases being used, one for metadata operations and one for the user scripts.

I like the changes you made to only connect once and want to keep that. But we need to figure out how best to specify the database for the metadata queries. Maybe the easiest thing we can do is just update the query templates to take a three part name and use all the parts of the `change_history_table` object?

MACKAT05:
if  a script calls  any of  the permutations of "Use/Create database/schema/warehouse X"  or "use role Y" we would have an issue since the state of the session would not return to configuration, and create unexpected behavior compared to legacy.  
The relevant info is embedded in conargs... and is embedded in the class and can be accessed in the same code as the set tag functions.   these can be expanded and renamed i'll push the update in a second.

MACKAT05:
I think that covers it... the metadata queries happen up front before any session changes occur and the ---ahhh i see... fixing that now...

MACKAT05:
I attempted to make adjustments to the tests to get  them to pass... currently the 24 of them are complaining about the same issue that I think might be a config issue...  since one looks like a linux path and the other a relative path... 
image
  the 25th i just fixed

MACKAT05:
sometimes all you need is a little sleep  ; all tests passing now :
image

sfc-gh-jhansen:
Hey there @MACKAT05, I can't see any of your new changes here, so I guess we do need a new PR for this. I thought you could add to a previously merged one but it doesn't look like we can.

MACKAT05:
making it now :)

PR #145, opened 2023-02-08
Added optional flag for isolating R and V scripts that will be grabbed during processing

CermitDFrog:
We're working with multiple branches in our development environment and when a dev is working on an R script. Other deployments will reset that R script to default.

I added an optional flag that allows us to isolate schemachange to work only on a comma delimited list of files, with the exception of A scripts which will still run always.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@ All notable changes to this project will be documented in this file.
 
 *The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).*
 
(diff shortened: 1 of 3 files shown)


joshpulse:
This would be very useful for us to have this capability added as well. +1 to this idea

sfc-gh-jhansen:
Hey there @CermitDFrog, this change seems reasonable. But we've just released some significant refactoring changes with version 3.5.0. Can you please add your changes to that refactored version and resubmit. And please don't forget to update the version as well as the README and CHANGELOG files.

CermitDFrog:
@sfc-gh-jhansen Thanks. Merged in master and updated README and CHANGELOG.

Thanks.

PR #147, opened 2023-02-08
Additional Bug Fixes

MACKAT05:
Updated loggin function to insert with fully qualified name;  added session normalising steps to reset ROLE; WAREHOUSE;DATABASE adjacent to the Session tag actions... removed errant variable in config dictionary that made it into the refactoring using dictionaries.  all other test issue asscociated with key names changing from kebab-case to snake_case addressed and fixed.


MACKAT05:
@jeremiahhansen is there a way to run these tests locally instead of through github actions?  i would have liked to have sorted out the issues in one commit.

sfc-gh-jhansen:
Alright @MACKAT05, I think we're just about there. I tested out these new fixes and found a few more bugs which I just committed. I'll do a bit more testing and push and updated package to PyPI by tomorrow hopefully.

I really like the refactoring changes you made and have been wanting to do that for a while (especially around not creating a new connection to Snowflake each time). But it's a big refactor so just want to make sure it doesn't change behavior for anyone.

Thanks again!

sfc-gh-tmathew:
awesome work @MACKAT05!

sfc-gh-jhansen:
Alright @MACKAT05 (and @sfc-gh-tmathew), just published 3.5.0 to PyPI. Great changes in this release, including the refactoring. I went through it again and don't think we should have any behavior changes that would affect users, but let's keep our fingers crossed for now 🤞. Thanks again.

PR #153, opened 2023-02-14
state that snowflake_database var does not alter where change history table is created

RichieBzzzt:
Minor alteration to the README to make it clear that using a two part name in the ```CHANGE_HISTORY_TABLE``` argument will result in the METADATA database being used to store the change history table, and not the value in the ```SNOWFLAKE_DATABASE``` argument.

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,7 +171,7 @@ The Jinja autoescaping feature is disabled in schemachange, this feature in Jinj
 
 ## Change History Table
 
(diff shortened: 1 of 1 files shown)


RichieBzzzt:
Happy to alter the wording on this. I assumed that the snowflake_database var would set the database being used in the context of the connection, as I'm used to SMO and ADO.NET, and probably other people used to those libraries would make the same assumption.

README.md
```diff
@@ -171,7 +171,7 @@ The Jinja autoescaping feature is disabled in schemachange, this feature in Jinj
...
+schemachange records all applied changes scripts to the change history table. By default schemachange will attempt to log all activities to the `METADATA.SCHEMACHANGE.CHANGE_HISTORY` table. The name and location of the change history table can be overriden by using the `-c` (or `--change-history-table`) parameter. The value passed to the parameter can have a one, two, or three part name (e.g. "TABLE_NAME", or "SCHEMA_NAME.TABLE_NAME", or "DATABASE_NAME.SCHEMA_NAME.TABLE_NAME"). This can be used to support multiple environments (dev, test, prod) or multiple subject areas within the same Snowflake account. If you use the two part name then it will attempt to create the change history table to the 'METADATA' table, and not the database set in the ```SNOWFLAKE_DATABASE``` variable. By default schemachange will not try to create the change history table, and will fail if the table does not exist.
```
sfc-gh-tmathew, line 174: `METADATA` Database instead of table.

sfc-gh-tmathew (changes requested):
@RichieBzzzt Please make the correction and resubmit for review.

Issue #161, opened 2023-02-24
Role, Database, and Warehouse Not Resetting Before Logging


nwea-yeti-jw:
We found that after running a migration, the Snowflake role, database, and warehouse are not resetting if those values changed within the migration. In other words, if the migration itself contains `use role my_role;` within the migration, the insert query for the log will still try to use that role rather than what was set in environment variables. This created an issue when we found that the role may not necessarily have privileges on the database where the logs are inserted.

In the current cli.py script on line 433, you can see that `self.reset_session()` is called, but it is not called again after executing the migration; therefore when line 450 that has the log execution is called, the query will use the last role, database, and warehouse if set within the migration. We have confirmed this in testing. 

The workaround we have implemented is reverting to pre-3.5.0 and this is successful. We have also submitted a small PR with a fix.

sfc-gh-jhansen:
These changes were deployed in v3.6.0, see PR #195

PR #162, opened 2023-02-24
Reset the session before logging

nwea-yeti-jw:
Issue #161 The session needs to be reset in case role, database, or warehouse were set within the migration script.

### DIFF
--- a/schemachange/cli.py
+++ b/schemachange/cli.py
@@ -436,6 +436,8 @@ class SnowflakeSchemachangeSession:
       self.reset_query_tag()
       end = time.time()
       execution_time = round(end - start)
+      # reset the session before logging
+      self.reset_session()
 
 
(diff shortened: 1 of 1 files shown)


pavlovecjakePFG:
This PR fixes a bug that was introduced in one of the recent versions. I recently attempted to upgrade to the newest version (current = 3.4.1) and had to roll back because of this bug. If a role is assumed in a script that does not have access to modify the change history table, then you get an error.

temyers:
Any update on this?

This caused a breaking change preventing upgrade from a previous version.

MaxKrog:
Bump! We're having the same issue and a merge of this fix would be highly appreciated.

Thanks for a great tool.

sfc-gh-jhansen:
These changes were deployed in v3.6.0, PR #195. Thanks everyone.

PR #172, opened 2023-04-25
Added Always First Functionality

### LINKED ISSUE #164: Add Always Execute First Functionality
### Description:
Add the functionality to execute specific scripts first on every execution of schemachange using the naming convention `F__DESCRIPTION.sql`.

### Benefit:
For use in CI/CD process:
1. Create clone of production environment to create the QA environment.
2. Apply permissions required for the QA environment.
3. Execute rest of schemachange workflow to apply deployment script to QA environment. If this fails, then the deployment would fail a production execution.

By tagging scripts to be executed first, schemachange can create the cloned environment and apply the appropriate permissions before it attempts to deploy any changes.

### Best Case Enhancement:
The functionality would be improved if an optional condition could be applied so it is only executed against QA and not to PROD, since it would not be needed against that environment. For instance, if the naming convention was `F_QA__DESCRIPTION.sql`, then it would only run against the QA environment. This could also be implemented with a command line argument like, `-af, --always-first` or in the line argument of `always-first: TRUE`.

CaptCorpMURICA:
This pull request resolves #164.
- This enables the new file type `AF__*.sql` which is processed before all other file types.
- This option defaults to `False` but can be enabled with the configuration `-af` or `--always-first`. This ensures that these files can be selectively executed depending on the environment.
  - e.g. In the CI/CD pipeline, you enable this configuration option when you need to clone the production environment to create the QA environment. However, this option is turned off for the elevation from QA to Prod because no cloning operation is taking place.
- Tests were added to ensure the functionality works as designed. These tests cove:
  - Returning the proper number of files executed when the option is turned on or off.
  - Returning expected errors based on naming conventions.
  - Validating the appropriate data for the Change History Table.
  - Validating the appropriate execution order (`AF->V->R->A`) when the option is turned on.
  - Validating the appropriate execution order (`V->R->A`) when the option is turned off.
- Added this new functionality to the README documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@ For the complete list of changes made to schemachange check out the [CHANGELOG](
    1. [Versioned Script Naming](#versioned-script-naming)
    1. [Repeatable Script Naming](#repeatable-script-naming)
    1. [Always Script Naming](#always-script-naming)
+   1. [Always First Script Naming](#always-first-script-naming)
    1. [Script Requirements](#script-requirements)
    1. [Using Variables in Scripts](#using-variables-in-scripts)
       1. [Secrets filtering](#secrets-filtering)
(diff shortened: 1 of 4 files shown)


CaptCorpMURICA:
Located a bug. Correcting and resubmitting pull request.

PR #182, opened 2023-06-23
add a couple gotchas around change scripts

afeld:
Not totally clear to me _why_ these are the case... Is it due to the connector version? The underlying API being called? Anyway, they tripped me up, so hoping this can save others some confusion.


sfc-gh-jhansen:
Included in the new 3.5.3 release. Thanks!

PR #188, opened 2023-07-20
add logic if password and okta authenticator are passed as args

AnilBakuru:
Adding logic to handle issue # https://github.com/Snowflake-Labs/schemachange/issues/181

PR #189, opened 2023-08-17
Disable password authentication when SNOWFLAKE_AUTHENTICATOR has a value.

MatthewFlood-BO:
When using `authenticator=externalbrowser`, the `snowflake.connector.connect()` method requires a value for the password parameter.  However, if we provide an value for the password in the env variable SNOWFLAKE_PASSWORD, then `cli.py` forces password authentication. This means it is currently impossible to use externalbrowser authentication.

I would argue that if we set a value for SNOWFLAKE_AUTHENTICATOR, then password authentication is explicitly unwanted.

### DIFF
--- a/schemachange/cli.py
+++ b/schemachange/cli.py
@@ -271,7 +271,7 @@ class SnowflakeSchemachangeSession:
         warnings.warn(_warn_password, DeprecationWarning)
         snowflake_password = os.getenv("SNOWSQL_PWD")
 
-    if snowflake_password:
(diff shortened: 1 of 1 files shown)


sfc-gh-tmathew:
@MatthewFlood-BO This change breaks if password authentication. Check the #188 fix and see if that works for you.

cc: @jeremiahhansen

sfc-gh-tmathew:
Thank you for your contribution @MatthewFlood-BO. We have merged the fix into v3.5.4.

PR #154, opened 2023-02-15
fix(readme): typo in auth table

danyalprout:

PR #192, opened 2023-09-01
Fix Authentication Workflow

sfc-gh-tmathew:
Including fixes reported through #154, #188 and #189. Authenticator workflow has been tested for all supported combinations.


sfc-gh-twhite (approved):
LGTM, all version values updated from 3.5.3 to 3.5.4. 

We are adding a new expected env variable of `SNOWFLAKE_AUTHENTICATOR` here.

In the future, we should remove checks for SNOWSQL prefixes.

PR #180, opened 2023-06-13
Add support for specifying the schema via config

podung:
## PR: Configure Snowflake Schema via Config

I'd like the ability to _**optionally**_ specify the `snowflake_schema` via config.

### Example of Problematic Script

```sql
CREATE OR REPLACE FUNCTION {{ db_name }}.{{ schema_name }}.Get_Customer_Name_By_ID(
  ID integer
)
RETURNS varchar
LANGUAGE SQL
as
$$
  SELECT FIRST_NAME || ' ' || LAST_NAME
  FROM CUSTOMER
  WHERE ID = ID
$$;
```

### Explanation
The `schemachange` docs specify
> schemachange is designed to be very lightweight and not impose to many limitations. Each change script can have any number of SQL statements within it and must supply the necessary context, like database and schema names

For most objects, it is sufficient to just declare the db name and schema name in the proc / function / table declaration.  However some objects (namely SQL UDTFs in this example) in snowflake seem to ensure the object can compile.  Since a schema is not provided on the `SELECT ... FROM CUSTOMER` statement, and no `USE SCHEMA` directive has been set, the error occurs.

### Work-Arounds
1. In the UDTF, a fully qualified name can be given (e.g. `Select ... FROM {{ db_name }}.{{ schema_name }}.CUSTOMER`).
    - This is less-than-desirable for our use case, as it causes the resultant object DDL to include the db/schema in the UDTF body.  We have a project requirement to keep object DDLs identical between databases (dev/test/prod/etc).

2. In the migration, before the `CREATE OR REPLACE` directive, we can simply add the statement `USE SCHEMA {{ schema_name }};`
    - This work-around is acceptable, but we have to reach for it in many of our migrations

### Why this change is useful
Our team has a requirement to keep the object DDLs identical in each environment (Dev, test, prod, etc).  There are a few instances where snowflake requires fully qualified object names to verify compilation during object creation (namely: SQL UDTFs).  Instead of having to add the `USE SCHEMA {{ schema_name }};` directive in many of our migration files, we'd like to just supply the schema up front and have `schemcachange` execute the statement for us at the beginning of the run.

### Tasks left before PR should be accepted
- [ ] Make `Snowflake_Schema` configuration optional (how it is currently coded it will blow up if not supplied)
- [ ] Ensure snowflake schema can be supplied via CLI as well
- [ ] Modify Documentation
- [ ] Add / modify any tests as necessary

If you are interested in this PR I would be happy to complete the above tasks.  Thanks for the library and for the consideration.

PR #196, opened 2023-09-14
Added Support for schema syncing via cli

yakshup:
Extending #180 with cli options and tracking in change history


sfc-gh-tmathew:
@yakshup By adding the schema column without the altering the table structure will result in failure. This will also require documentation updates and ensure that existing users of schemachange do not loose the change history when upgrading to a release that includes this change. This cannot be accepted in its current state and needs more work.

podung:
Sorry all - I didn't see this PR hanging off my original.  I finally have my original fully up to date and ready to commit, though I needed to open a new PR: #210 .

It has tests, documentation, cli support, and yml config support

podung:
On the additional specific effort in #196, I'm not exactly clear what was being attempted.  @yakshup can you elaborate on what use case you were trying to support by wanting to track a schema column on the change history table?

yakshup:
> On the additional specific effort in #196, I'm not exactly clear what was being attempted. @yakshup can you elaborate on what use case you were trying to support by wanting to track a schema column on the change history table?

@podung We had a requirement to track changes per schema on which changes are made. I believe this was something which was raised in #180.

sfc-gh-tmathew:
Thank you for contributions but we will not be including this change to the schemachange feature.

Issue #215, opened 2024-01-02
SQL Syntax error when Snowflake role name includes special characters such as hyphens


jonathancisneros:
The **USE ROLE** and the other USE query template strings here:
https://github.com/Snowflake-Labs/schemachange/blob/master/schemachange/cli.py#L213

are not using quoted values for the placeholder values, if there are special characters in the name of those values, the execution of the **reset_session** function fails with SQL compilation or syntax error.

please update the USE xxx template strings to support special characters by using quoted values like this:
_q_set_sess_role = 'USE ROLE "{role}";'

dwelden:
Duplicate of #186?

sfc-gh-tmathew:
Thank you for highlighting this gap @dwelden @jonathancisneros 

Just released 3.7.0 that addressed roles with special characters. Please test it out and let us know if this issue can be closed.

dwelden:
This issue is resolved.

sfc-gh-tmathew:
Thank you @dwelden for confirming. Closing this issue for now.

PR #216, opened 2024-01-02
Support role, database, schema or warehouse names with special chars

jonathancisneros:
Fix for issue https://github.com/Snowflake-Labs/schemachange/issues/215


sfc-gh-tmathew (changes requested):
please use identifier function instead of the double quotes because snowflake already has documented support for objects with special characters.

PR #217, opened 2024-01-10
Format with black

zanebclark:
I've added the black formatter to the pre-commit hooks along with the flake8 linter. I addressed the issues raised by these tools along with some formatting and type hint errors. The functionality of the script should be the same. The development experience should be a bit nicer.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -13,7 +13,6 @@ Choose the appropriate issue template to contribute to the repository.
 
 Thanks for your interest in contributing code to schemachange!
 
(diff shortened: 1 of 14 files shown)


sfc-gh-twhite:
Thank you for opening this, @zanebclark! I apologize, we have been behind on many of these PRs and Issues and are trying to catch up. 

Many of these formatting changes had been on our roadmap and have been taken care of elsewhere since then. I do think that there are portions of this we would like to use, particularly exploring flake8 and some of the type annotation work; once we fold in some of the other upstream changes, I would like to include this one as well.

PR #218, opened 2024-01-10
Package with Poetry

zanebclark:
Migrate to Poetry for packaging. This simplifies dependency maintenance and environment setup. Given that I don't have the pypi api key, I was unable to validate GitHub Action

Issue #205, opened 2023-10-26
Dependency lock for Pandas (and arguably all packages) is too strict


dwreeves:
TLDR: I believe Pandas should either be refactored out of the library entirely, or the upper bound on the version should be removed. There is also a very strong argument for removing the other upper bounds.

Upper bounds on dependencies are well-intended in practice to provide assistance to users doing standalone installations, but they can also have the opposite effect of making installation in more complex environments a burden.

---

The `setup.cfg` has the following dependencies:

```ini
install_requires =
    jinja2~=3.0
    pandas~=1.3
    pyyaml~=6.0
    snowflake-connector-python>=2.8,<4.0
```

All of these strike me as a little too strict, given that only core API behavior is being utilized for each of the packages.

Pandas is the most noteworthy. Pandas is hardly being used in the code in the first place. (In fact, the code could be refactored so that it is not even a dependency.) What little is being used is core API behavior. And, unlike pyyaml and jinja2, there exists a major version update (Pandas 2.x) that has a semantic version which is incompatible with the upper bound of `<2`. There needn't be any version locking whatsoever for this.

---

While we are at it:

- PyYaml is also hardly being used; just the core API, which is not subject to any serious change.
- Jinja2 is a core part of the library, but it has a very stable API. Major version bumps have historically had very minor API implications and are consistently migrated very smoothly.

Overall these upper bounds strike me as overly cautious, especially in the case of Pandas.

StLWallace:
Came here to post the same issue.

For local testing, we're using a monorepo where we also run Snowpark, which if I'm not mistaken requires pandas 2.2.1. I agree that the pandas version here should be at least upgraded, and ideally removed.

sfc-gh-tmathew:
Hello @StLWallace @dwreeves 

v3.7.0 has removed pandas dependency. Please check and let us know.

dwreeves:
Thanks @sfc-gh-tmathew! Do you have any thoughts on the additional dependency restrictions I mentioned may be too strict, specifically PyYaml and Jinja2?

sfc-gh-tmathew:
@dwreeves We have a thread going on about organizing the code to make it easier to test, extend. The idea was to avoid breaking a tested release by pinning the highest version we tested with. Will test later versions of PyYaml and Jinja2 to make the upper limit less concerning.

dwreeves:
Ok.

[...]

PR #222, opened 2024-01-22
Flag for error on ignored versioned migration and versioned script regex

zanebclark:
This pull request started out as a feature addition. It ended up being a bit of a repo refactor. 
- Format with Black. There's already an open PR on this: https://github.com/Snowflake-Labs/schemachange/pull/217
- Package with Poetry. There's already an open PR on this: https://github.com/Snowflake-Labs/schemachange/pull/218
- If the "change_history_table" string is a schema and a table, use the "snowflake_database" instead of "METADATA"
- Deprecate `verbse` cli argument. Opt for `log-level` instead
- Replace print statements and `if verbose` checks with a [structlog](https://www.structlog.org/en/stable/) logger and log levels.
- Removed SecretManager in favor of a structlog processor that implements very similar logic. 
- Remove pandas dependency in favor of a dictionary. This should address one of the complaints in https://github.com/Snowflake-Labs/schemachange/issues/205. 
- Centralize config-related logic and input checking into a `Config` object. For example, there was logic for checking the validity of a directory and the presence of environment variables. Centralizing this logic is conceptually easier for me to wrap my head around and makes testing much easier. 
- Generally, moved a bunch of class definitions out of `cli.py` and into separate files
- Stored script details in a `Script` class to clarify the API
- Removed a bunch of credential-related logic from `SnowflakeClient` and created a `Credential` factory. 
- Generally expanded test case coverage. 
- Added two CLI arguments: 
  - **raise-exception-on-ignored-version-script**: At my shop, we're using epoch time as our version number. If you create a script with an epoch time and your colleague merges their changes in before you get a chance to, the `max_published_version` will cause `schemachange` will skip your script without complaint. To address this, I handle versioned scripts almost identically to repeatable scripts. I store the metadata on the script's execution in a dictionary and compare that to the scripts in the script directory. If a script is "stale", hasn't been applied, and the flag is true: I raise an exception. 
  - **version-number-validation-regex**:  Again, we're using epoch time as the version number in my shop. If someone deploys a script prefixed with `Vspam__`, future scripts versioned with the current epoch time will be skipped. The `max_published_version` is now `spam` and cannot be trumped by a number. To address this, I've added a `version_number_validation_regex` CLI argument that will reject `spam` as a version number if I've specificed `\d{10}`

Not everything is peachy: 
- I think I might have broken some stuff around the `render` command. Specifically, the bit where we [clear config values for the render subcommand](https://github.com/Snowflake-Labs/schemachange/blob/master/schemachange/cli.py#L863). I'd appreciate some feedback on how to handle this. 
- I ran out of steam when I got to the `render.py` and `SnowflakeSession.py` modules. There's work to do there in terms of dependency inversion and testing. I'm putting this out there now for feedback, not because I think it's "done". Don't be surprised if this lack is fixed in the near future on this PR. 
- I saw the dependency checker on the pipeline. If that's a serious sensitivity, this PR might be dead in the water. 
- Performance likely took a hit, though I haven't done any profiling. I'm not too concerned about the performance of this utility, but someone here might be. 

Sorry for the monstrous pull request. Consider it a reflection of my inability to understand code that I didn't write :blush:


sfc-gh-tmathew:
@zanebclark Amazing work here. Will need to review and see if we can carve out the changes into multiple pull requests to be released without breaking previous patterns. For breaking changes, we will have to place them in a 4.0 release and plan for the future with some communications related to it. 

Give us some time to review but if you can carve out your PRs into categories that will be incremental that would be really helpful.

zanebclark:
@sfc-gh-tmathew , apologies for the delay in my response. I'm embarrassed to have created such a behemoth. I've been eating my own dogfood by using this branch to create and maintain Snowflake environments. I've encountered two bugs: 
1.  A dry run without a `change_history` table will result in a failure. I didn't properly mimic the previous logic to handle this case. 
2. When I configured the regex pattern as `\d{9}`, it failed to raise an exception when a microsecond epoch time was used as the version number. 

I included the conversion to Poetry for my own convenience. If there's no interest in simplifying package management, that bit can be stripped out. Let me know what you think.

sfc-gh-tmathew:
> @sfc-gh-tmathew , apologies for the delay in my response. I'm embarrassed to have created such a behemoth. I've been eating my own dogfood by using this branch to create and maintain Snowflake environments. I've encountered two bugs:
> 
> 1. A dry run without a `change_history` table will result in a failure. I didn't properly mimic the previous logic to handle this case.
> 2. When I configured the regex pattern as `\d{9}`, it failed to raise an exception when a microsecond epoch time was used as the version number.
> 
> I included the conversion to Poetry for my own convenience. If there's no interest in simplifying package management, that bit can be stripped out. Let me know what you think.

Please strip out the poetry part for package management. We will still need to review the rest.

sfc-gh-tmathew:
> If the "change_history_table" string is a schema and a table, use the "snowflake_database" instead of "METADATA"

This will be a breaking change to existing end users for those users who created METADATA database OR uses fully qualified change history table.

If you are going with a change history table per database, I can see why you need this option to default to this option. 

Please consider calling schemachange for each database and fully qualify the change history table with the database to override the metadata database reference.

zanebclark:
@sfc-gh-tmathew, I'm going to close this stale pull request and open a new one that isn't based on the Poetry packaging PR. I'll remove the breaking change you caught in that PR

sfc-gh-tmathew:
@zanebclark Thank you for your contributions and understanding.

PR #226, opened 2024-02-04
Add in repeatable migration batching process

nickpileggi947:
Add in repeatable migration batching process. This allows for repeatable migrations to be batched together to speed up deployment. The number of migrations batched together can be controlled via the config file and params. In the event of a failure in one of the batched repeatable migrations the repeatable migrations will be retried until the affected migration is found.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,517 +1,1018 @@
-# schemachange
-<img src="images/schemachange-logo-title.png" alt="schemachange" title="schemachange logo" width="600" />
-
-*Looking for snowchange? You've found the right spot. snowchange has been renamed to schemachange.*
-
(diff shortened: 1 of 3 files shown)


nickpileggi947:
Any feedback on this/any desire to merge this in if I resolve conflicts from the latest changes

sfc-gh-tmathew:
Hi @nickpileggi947,

Thank you for this PR to add repeatable migration batching for faster deployments.

This PR has been open for about 2 years and likely has significant conflicts. The batching concept is interesting, but we'd need to understand:

1. What performance improvement did you observe with batching?
2. How does the retry-on-failure logic work when a batch fails?

**Questions**:
1. Are you still interested in pursuing this feature?
2. If so, would you be able to rebase against `master`?

If we don't hear back in the next few weeks, we may close this PR to keep our backlog manageable. You're welcome to reopen later.

Thanks!

PR #236, opened 2024-03-13
Preparing release mgmt changes

sfc-gh-tmathew:
Creating a draft pull request to include the improvements needed for release management.


IndexSeek:
This is really good stuff @sfc-gh-tmathew! I think we can get away without having the `ruff_cache` folder added to the gitignore as it's contents are already ignored, but that's just a small nitpick.

PR #238, opened 2024-03-20
Redact multiline secrets

rwberendsen:
**Problem**
When given a multi-line secret in the schemachange-config YAML, it would not be redacted, because just before calling `SecretManager.global_redact`, the config vars section would be serialised as YAML, adding leading whitespaces to the multiline secret

**Proposed solution**
Recurse over the config vars section, and redact all strings in it, and only then serialise as YAML.  When redacting multiline secrets, preserve newlines.

**Minor edits added in**
- The method that was extracting secrets would crash if the value was of a type that did not have the `strip()` method. Added an explicit raise of a ValueError if a secret is not of type `str`
- Added tests for added functionality

**Notes**
For redacting in SQL queries nothing was changed, as indenting while rendering Jinja templates is under control of the user of schemachange.

### DIFF
--- a/tests/test_redact_config_vars.py
+++ b/tests/test_redact_config_vars.py
@@ -1,0 +1,61 @@
+import pytest
+from schemachange.cli import redact_config_vars, SecretManager
+
+
+def test_redact_config_vars_given_secret_val_should_redact_it():
+    config_vars = {'secret' : 'secret_val'}
+    sm = SecretManager()
+    SecretManager.set_global_manager(sm)
+    sm.add('secret_val')
+    results = redact_config_vars(config_vars)
+    redacted = {'secret' : '**********'}
+    assert results == redacted
+
+
+def test_redact_config_vars_given_number_like_secret_should_redact_it():
+    config_vars = {'secret' : '123'}
+    sm = SecretManager()
+    SecretManager.set_global_manager(sm)
+    sm.add('123')
+    results = redact_config_vars(config_vars)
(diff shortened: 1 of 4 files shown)


tests/test_redact_config_vars.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+def test_redact_config_vars_given_secret_should_redact_it_in_any_string_it_finds_it_even_in_lists():
+    # At the moment, it's not documented that we could specify lists as values for variables, but,
+    # nothing is preventing people from trying things. Just in case, redact_config_vars is
+    # able to also redact strings inside lists. That's why this test is here.
+    # Finally note that tuples are not scanned, just making the point that this redact function cannot
+    # do everything; YAML deserialises lists as Python lists, not tuples, so we have no use case for tuples.
+    config_vars = {'secrets': {'password' : 'hi'}, 'jdbc': 'jdbc:mysql://127.0.0.1',
+                   'accounts': [{'user': 'john', 'password': '**'}], 'greetings': ['hi', 'hello'], 'greetings_tuple': ('hi',)}
```
rwberendsen, line 52: accidentally already masked `hi` already once in this dict :-)

rwberendsen: Fixed

sfc-gh-tmathew:
@rwberendsen We are catching up back on keeping tabs. Could you resolve the conflicts in this PR?

Thank you for your contribution. Apologies for the delay in responding.

PR #242, opened 2024-04-04
Updating Pytest criteria

sfc-gh-tmathew:


sfc-gh-tmathew:
@sfc-gh-twhite Added dev and master branch pytest github actions. We have dependency reviews triggered for each PR. Will pull this to lower branches once merged.

.github/workflows/dev-pytest.yml
```diff
@@ -0,0 +1,32 @@
...
+
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.8", "3.9", "3.10", "3.11"]
```
sfc-gh-twhite, line 16: For dev, let's pick one version of Python for now. This should speed up tests where only simple changes are occurring. 

```
    steps:
      - uses: actions/checkout@v4
      - name: Set up Python 3.11
        uses: actions/setup-python@v4
        with:
          python-version: "3.11"
```

sfc-gh-twhite (changes requested):
I think one change, but this looks great!

sfc-gh-twhite:
And should we add a workflow_dispatch so we can trigger these ad-hoc?

sfc-gh-tmathew:
> And should we add a workflow_dispatch so we can trigger these ad-hoc?

Added workflow dispatch for manual workflow against Dev

sfc-gh-tmathew:
@sfc-gh-twhite Added the requested changes.

sfc-gh-twhite (approved):

PR #245, opened 2024-04-04
Log Current Session ID

opodoprigora:
Log Current Session ID


opodoprigora:
@sfc-gh-twhite please take a look when ever you have a chance.

sfc-gh-twhite:
Thank you for the PR!

I would opt to use the session_id that is available from the property of the Snowflake connection. For example, we can do this:

```py
import snowflake.connector

con = snowflake.connector.connect()

print(con.session_id)
```
I believe that avoids the extra query submission. :)

opodoprigora:
@sfc-gh-twhite good call 💪 , updated.

sfc-gh-twhite:
Awesome job! We're doing a good bit of overhauling from a maintenance perspective of this project, but hopefully we can get this one in soon.

PR #259, opened 2024-07-02
Removed Pandas as a dependency

naruvimama:
Replaced the small piece of code that uses Pandas, removed it as a dependency.


sfc-gh-twhite:
Thank you for your contribution, @naruvimama! 

We plan to implement your changes in an upcoming release once we have a bit more time to test them alongside some of the other changes that are in flight.

PR #271, opened 2024-08-01
Comply with pre-commit

zanebclark:
- Expand CONTRIBUTING.md documentation to include "editable" package installation and pytest execution
- Provide argument to "name-tests-test" pre-commit hook that matches the pytest naming convention
- Quote Jinja expansions to comply with "check-yaml" pre-commit hook
- Add pre-commit to dev requirements


sfc-gh-twhite (approved):
LGTM! Thank you for doing this, very good catch as well adding pre-commit to the dev extra.

PR #282, opened 2024-09-17
New Baselining and Script Dependency Generation Features

sanelson:
Hello,

Thank you all for creating and maintaining such an awesome tool. It has been very helpful as we mature our Snowflake DB management processes.

TLDR; Added some new Schemachange features and wanted to share.

The two main features I added are:

## Baselining subcommand

I had several pre-existing Snowflake DB's that I wished to migrate to using Schemachange. I needed a way to set the current version of the DB and create a simulated change history to match up with the schema scripts in my repo. Adding this feature was easier than creating the change history from scratch. Hopefully it will be useful for someone else. Instructions on typical usage are in the README.

## Script Dependency Generator

I track my Snowflake Dynamic Tables and Views in R__ scripts and needed a way to re-run scripts which had parent dependencies change underneath them. Using SQLGlot, I parse the SQL in all R__ scripts and identify statements which are CREATING or ALTERING DB objects and then determine which parent objects (table, view, dynamic table) a child (view, dynamic table) object depends on. Using this Parent/Child relationship, I use NetworkX to generate a directed graph of all known DB objects and their associated Schemachange scripts. Any scripts which have changed since the last deployment (new V/R, changed R) are analyzed to see what downstream dependencies/scripts are affected and all of those scripts are re-run. This feature is controlled through the `--script-deps` option to the `deploy` subcommand. This feature needs a bit of improvement and my assumptions are probably a bit naive in how I implemented it. But it works great for our team so far.

## Misc smaller features

- Added support for Rich formatted output
- Added SQL formatting option to render. 
- Added a '--silent' rendering option to allow passing of raw SQL to other processes or files via STDOUT without logging.
- Migrated from print statements logging to Python logging module
- Refactored argument parsing configuration to share common args between similar tasks (deploy, baseline)
- Fixed issue with backwards compatible deploy subcommand not supporting main script help args (-h, --help)

Apologies for the size of this pull request. If it would be better for me to break this up into smaller chunks, I'm happy to try doing that.

Thanks!

Sam Nelson


zanebclark:
@sanelson, the script dependency generator sounds impressive. I don't have much experience with dynamic tables and repeatable migrations. Can you help me understand the need to execute parent or child scripts? You lost me with the term "parent dependencies" 🙃. 

I've noted the need for dependency tracing with view definitions. Assume you define all of your views as repeatable migrations and execute them alphabetically (the default). When creating a new environment, you might be trying to create a view that references a view that hasn't been created yet.
- `R__A_View_depends_on_B_View.sql`
- `R__B_View.sql` 

My low-tech hack is to introduce a "version" after the repeatable migration indicator to control the order of application: 
- `R__0001_B_View.sql` 
- `R__0002_A_View_depends_on_B_View.sql`

I think the script dependency generator would solve this problem without explicit filename changes. 

Unfortunately, it also introduces a bunch of additional complexity into a relatively simple tool:
- If you define multiple objects in a file containing a dependent migration, will the entire migration be applied again? Or, would the specific dependency be applied, ignoring the other unrelated objects? 
- Would versioned scripts also be re-applied? 
- If so, how would we keep track of the application history there, as the currently model only supports the metadata around a single versioned migration application? 
- How do we handle situations where an object is altered in multiple migrations? 

I've only briefly reviewed your solution. You may have addressed many of the questions above.

sanelson:
> @sanelson, I love the baselining feature! It would support another feature I've considered:

I think that your "checksum check" feature would be really useful and not difficult to implement. I was considering adding an Audit feature to look for schema "drift" and this would be a great feature to develop alongside it. Should be relatively straightforward to add to the baseline feature.

Might be nice to add a WARNING level log message if checksum drift is detected. If your flag is enabled then the program would throw an ERROR and exit. Thoughts?

Related to version checksum checking, the [SQLGlot library](https://github.com/tobymao/sqlglot/tree/main) that I used for the dependency generator [first tokenizes the SQL](https://github.com/tobymao/sqlglot/blob/main/posts/python_sql_engine.md#parsing) that it is parsing before doing all the magic to build an [AST](https://github.com/tobymao/sqlglot/blob/main/posts/ast_primer.md). I leverage a number of Jinja templates to build my final SQL scripts and I've been worried about inadvertent white space changes in shared templates causing unwanted updates to Repeatable scripts. By using only the tokenized SQL from each file to generate a checksum, I think this issue could be avoided. It could also do neat stuff like ignore comments, etc. The downside to using SQLGlot so far seems to be idiosyncrasies with certain SQL patterns that are valid in Snowflake but cause parsing failures in SQLGlot (TABLE X UNION ALL TABLE Y is a good example). I would want to write some robust unit tests to catch these various conditions and ideally push fixes upstream to the SQLGlot project to fix them.

sanelson:
Re: Dependency Generator

### TLDR; 

I whole-heartedly agree that this feature introduces a whole bunch of additional complexity in a simple and easy to use tool where it probably doesn't belong... With the awesome work that you've done refactoring, this functionality could easily be moved to a separate utility/helper/plugin script and not pollute the main tool/script.

### Dependency Generator Background

@zanebclark, the script dependency generator was really just developed to speed up the Schemachange runs when updating dynamic tables. Some of our dynamic tables take minutes to rebuild everytime they are recreated and I was trying to avoid having to run all of these scripts every time we did a deploy. Originally we had all of our dynamic table definitions as R__ scripts but ran into an issue when one dynamic table was updated (checksum changed) but then another dynamic table which depended on the updated dynamic table was not updated (no checksum change). This caused subsequent failures in Snowflake when querying the "downstream or child" dynamic table. Honestly I probably should have invented a new file prefix (D__ ?) for this feature as I think overloading the R__ scripts with this functionality just creates confusion and complexity.

> My low-tech hack is to introduce a "version" after the repeatable migration indicator to control the order of application:
>     * `R__0001_B_View.sql`
>     * `R__0002_A_View_depends_on_B_View.sql`

This is exactly the solution we have used too! I've only been using the dependency generator to trigger the execution of these ordered scripts regardless of checksum changes. I suppose that a much simpler alternative to using my dependency generator would be to re-run all repeatable scripts with a version number greater than that of the one changed. The downside to this is that it depends on the human element and requires people to correctly analyze and name all the dependencies.



>     * If you define multiple objects in a file containing a dependent migration, will the entire migration be applied again? Or, would the specific dependency be applied, ignoring the other unrelated objects?

Very simplistic right now, if any object changes in a file then that whole file/migration will be applied again.

>     * Would versioned scripts also be re-applied?

Nope, this was only targeted to R__ scripts. Basically just another checkpoint to see if a script needed to be re-run in addition to checksum changes.

>     * How do we handle situations where an object is altered in multiple migrations?

Great question! This is where using a directed graph made the dependency analysis so much easier. I basically build a graph of ALL scripts and their objects. This graph includes version scripts as well. Then if any script (and it's objects) are changed/added, a dependency tree is generated for each of those objects. This tree consists of the object definition as well as the migration file(s) it is defined in. Now that I'm thinking about it, I need to double-check my code to ensure that a V__ script couldn't be inadvertently re-executed.

zanebclark:
> Might be nice to add a WARNING level log message if checksum drift is detected. If your flag is enabled then the program would throw an ERROR and exit. Thoughts?

We're almost there already. Check [this line](https://github.com/zanebclark/schemachange/blob/feature/code-reorganization/schemachange/deploy.py#L119) out. Not much more to do than add a flag and raise the exception accordingly. Teamwork = dreamwork!

zanebclark:
> Originally we had all of our dynamic table definitions as R__ scripts but ran into an issue when one dynamic table was updated (checksum changed) but then another dynamic table which depended on the updated dynamic table was not updated (no checksum change). This caused subsequent failures in Snowflake when querying the "downstream or child" dynamic table.

Please excuse my dynamic table ignorance. If a child dynamic table is updated, the parent dynamic table must also be updated? Using the analog of a view, I can understand that a column name change or the addition of a new column in the child view would cause the parent view to fail, but wouldn't you also want to update the parent view in that case? I imagine you're not talking about column names here. Can you help me understand this interdependency?

sanelson:
> Please excuse my dynamic table ignorance. If a child dynamic table is updated, the parent dynamic table must also be updated? Using the analog of a view, I can understand that a column name change or the addition of a new column in the child view would cause the parent view to fail, but wouldn't you also want to update the parent view in that case? I imagine you're not talking about column names here. Can you help me understand this interdependency?

I think my wording was really bad. Sorry about that! Let me try writing up an example and drawing a picture instead...

Let's say you have 4 dynamic tables, **A**, **B**, **C** and **D**. **A** isn't used by any other dynamic tables. However, dynamic tables **B** and **C** are used in the definition of **D**.

# A
```SQL
CREATE OR REPLACE DYNAMIC TABLE A
  AS
    SELECT 1 as mycol;
```

# B
```SQL
CREATE OR REPLACE DYNAMIC TABLE B
  AS
    SELECT 1 as anothercol;
```

# C
```SQL
CREATE OR REPLACE DYNAMIC TABLE C
  AS
    SELECT 1 as lastcol;
```

# D
```SQL
CREATE OR REPLACE DYNAMIC TABLE D
  AS
    SELECT anothercol as newcol FROM B
    UNION
    SELECT lastcol as newcol FROM C;
```

```mermaid
graph TD;
    A;
    B-->D;
    C-->D;
```

Changes to table **A** won't cause any downstream issues. However, if certain things are changed about the definition of dynamic tables **B** or **C**, then dynamic table **D** must be re-created, otherwise you may get errors in Snowflake when querying dynamic table **D**. I don't know the criteria of what kind of changes require recreation, but it has created issues for our team multiple times. Here's a [Reddit post](https://www.reddit.com/r/snowflake/comments/180ki8n/management_of_dynamic_table_changes/) that might do a better job of explaining the issue we were running into.

Hope that helps!

zanebclark:
Thanks for the additional information. From the reddit post: 
> No, in my case I had an existing varchar and adjusted the case statement that changed how the value was calculated. No change to the number, type, or names of columns. That caused a DT that depended on it to fail (and a DT following that one as well)... so a string of cascading failures. I had to just run the existing CREATE DYNAMIC TABLE... for all downstream DTs to fix them.

I am very surprised by this behavior. It feels more like a bug in Dynamic Tables than anything. I wasn't able to find anything in the documentation detailing which scenarios would require an update to the chain of tables. I'd also be interested in hearing from Snowflake support on the issue. It makes me leery of Dynamic Tables. 

The pull request is great work and sounds an elegant way to overcome this dependency. I'm also concerned about how far this feature deviates from the initial flywheel-like behavior. I'm not a maintainer, so my thoughts on that are worth what you paid for them 😉 . I wonder if there's room for a plugin ecosystem here? For example, consider [structlog](https://www.structlog.org/en/stable/).

- Schemachange gathers scripts and creates script objects. The default application rules would be applied to the script objects, with each object containing `apply: boolean`. The order of script applications would be encoded somehow as well.
- Script objects are passed to pre-plugins
- Schemachange executes scripts
- Script objects are passed to post-plugins

The script dependency generation feature could be a pre-plugin. It could alter the order of application and the `apply: boolean` property before the scripts are passed back to Schemachange to execute. For that matter, the default rules could be a pre-plugin as well. There's room for post-plugins that validate the health of views and dynamic tables. Again, scope creep from Flywheel behavior, but valuable for some.

sanelson:
Closing this pull request as a good deal of the functionality has been already implemented in the pull request #285 by @zanebclark. Will refactor remaining changes on top of new 4.0 release and submit new pull requests. Thanks!

PR #284, opened 2024-09-25
update schemachange to use master token and allow empty snowflake user, role, and warehouse

JianFangAtRai:
port changes from v3.6.1-oauth-file


JianFangAtRai:
not intended to push back to upstream

PR #303, opened 2024-11-22
fix: incorporating forked changes along with github actions updates

sfc-gh-tmathew:


populateConnection.sh
sfc-gh-twhite, line 1: This is a nitpick, and is not necessary, but I think we could make this all a single echo:

```bash
cat <<EOF > ./connections.toml
[default]
account = "${SNOWFLAKE_ACCOUNT}"
user = "${SNOWFLAKE_USER}"
role = "${SNOWFLAKE_ROLE}"
warehouse = "${SNOWFLAKE_WAREHOUSE}"
database = "${SNOWFLAKE_DATABASE}"
password = "${SNOWFLAKE_PASSWORD}"
EOF

echo "cat connections.toml"
cat ./connections.toml
```

This would require some testing, and if it isn't broken, I say probably don't fix it. :)

sfc-gh-twhite (approved):
This LGTM! Thank you @sfc-gh-tmathew and @zanebclark for the excellent work here!

populateConnection.sh
sfc-gh-tmathew: Good feedback. Will incorporate into future github actions updates.

faisalgen (commented):
after this update I m getting “password is empty error “ in my circle ci pipeline. till 20th Nov it was working fine from 25nov I m seeing this error.
Im using cloud formation template and not terraform.

faisalgen:
@sfc-gh-twhite @sfc-gh-tmathew  @jeremiahhansen @zanebclark . Need your help. After this update my  pipe is giving password is empty error and I m unable to resolve it. I m using yml config and toml file for connection.

PR #313, opened 2025-01-10
Fix/make get scripts case insensitive

zanebclark:


zanebclark:
@sfc-gh-jhansen @sfc-gh-tmathew @sfc-gh-twhite , another bug fix

sfc-gh-twhite (approved):
This LGTM! Thanks for working on this. Most of this looks like formatting changes as well. 

I'll let @sfc-gh-tmathew take a look since he's been a bit closer to the project than I have, but I'm good with this. Straight to master as well. I think we could bundle this with the other one and maybe do a 4.0.1 release or something.

zroytman:
Hi, any expectation when is it going to be released so we could use the fixed version? Thanks.

zroytman:
Hi, any updates regarding releasing the change? When can we see the solution? Thanks a lot.

zanebclark:
Thanks for merging this, @sfc-gh-twhite !

PR #332, opened 2025-05-16
feat: add py.typed marker for mypy users

fozcodes:
This lib is pretty well typed, but MyPy doesn't know to look for the types by default. Adding a py.typed file will tell MyPy to look at the type hints in the package directly. It's a pretty simple add for some big gains!

https://mypy.readthedocs.io/en/stable/installed_packages.html#creating-pep-561-compatible-packages

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -40,3 +40,6 @@ dev =
     pre-commit
     pyfakefs
     pytest
+
+[options.package_data]
+schemachange = py.typed


fozcodes:
Also, I've never published with `setup.cfg` so I have no idea if this is how you configure it. I think it is, but please let me know if not!

sfc-gh-twhite (approved):
I agree, this is a good add. Thank you!

At some point we will probably shift to a pyproject.toml approach. :)

PR #357, opened 2025-10-04
fix: support NO_COLOR env var for disabling colored output

### LINKED ISSUE #335: How do I disable the schemachange colored output?
** What are you trying to accomplish? **
Hi
I'm trying to migrate from schemachange 3.7.0 to 4.0.1 and my deployment tool is having problems parsing the ascii color escape codes produced in the output. It prints them which obfuscates the log (see screenshot below)
Image

** What options have you tried so far ? **
I've found that schemachange uses structlog, so I expected that setting environment variable NO_COLOR=1 would disable the color output as described in https://www.structlog.org/en/stable/console-output.html
However that does not work, schemachange still produces the color escape codes.
How to disable them?

sfc-gh-twhite:
Resolves #335

### DIFF
--- a/schemachange/__init__.py
+++ b/schemachange/__init__.py
@@ -1,6 +1,20 @@
+# Copyright 2025 Snowflake Inc.
+# SPDX-License-Identifier: Apache-2.0
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


sfc-gh-tmathew (approved):
Makes sense

PR #360, opened 2025-10-04
ci: migrate to uv based management

sfc-gh-twhite:
Let's go to uv and better pin these dependencies.

This let's us close out https://github.com/Snowflake-Labs/schemachange/discussions/343 as well.


sfc-gh-tmathew (approved):
Makes sense

Issue #394, opened 2025-11-25
Error moving from 4.0.1 -> 4.1.0 using private_key_file_pwd


enzoaccelerator:
**Description**  
Running schemachange 4.1.0  fails with `Password was not given but private key is encrypted`. But it works with version 4.0.1.

I'm using the following `connections.toml`
```
[main]
authenticator = "SNOWFLAKE_JWT"
account = "your_account"
user = "your_user"
role = "your_role"
warehouse = "your_warehouse"
database = "your_database"
private_key_file = "./private_key.pem"
private_key_file_pwd = "your_private_key_passphrase"
```

**Screenshots**  
<img width="1078" height="58" alt="Image" src="https://github.com/user-attachments/assets/c8329fa6-36c8-419c-86d9-0ef25d254078" />

```text
2025-11-20T17:44:44.108217Z [info     ] Using variables ...
2025-11-20T17:44:44.111181Z [error    ] Unexpected error error=Password was not given but private key is encrypted
Error: Process completed with exit code 1.
```

sfc-gh-tmathew:
Will be fixed in v4.2.0 (mid December Release). The private_key_file_pwd parameter will be correctly extracted from connections.toml. Both new names (private_key_file, private_key_file_pwd) and legacy names (private_key_path, private_key_passphrase) are supported until notified otherwise.

mibelbahri:
up, since PR is merged, can we please release 4.1.1 to patch to bug ?

mibelbahri:
@sfc-gh-tmathew is a patch release possible before the 4.2.0 ?

sfc-gh-tmathew:
@mibelbahri @enzoaccelerator Please verify using v4.2.0 and close the issue if you see it resolved.

PR #395, opened 2025-11-25
fix: re-introduce private_key_file_pwd in params for connection to snowflake, was removed from 4.0.1 to 4.1.0

enzoaccelerator:
Add private_key_file_pwd handling back to SnowflakeSession.

This pull request introduces a small update to the SnowflakeSession initialization logic. The change adds support for passing a private_key_file_pwd parameter when creating a session, allowing encrypted private key authentication to function again.

In 4.1.0 all connection parameters (from CLI, env, YAML, and connections.toml) are merged before SnowflakeSession is created, and DeployConfig.get_session_kwargs() already places both private_key_file and private_key_file_pwd in the kwargs handed to the session. 

However, the refactor that removed connection_name/connections_file_path from SnowflakeSession also rewrote the explicit_params dict that feeds snowflake.connector.connect(), and that new dict only forwarded private_key_file and the private_key_file_pwd entry was dropped. 

Because of that the connector now sees an encrypted key without its passphrase and fails with “Password was not given but private key is encrypted.”
Restoring private_key_file_pwd in SnowflakeSession’s connect_kwargs keeps the rest of the 4.1.0 refactor intact while making key‑pair authentication work again.

Will solve : https://github.com/Snowflake-Labs/schemachange/issues/394

### DIFF
--- a/schemachange/session/SnowflakeSession.py
+++ b/schemachange/session/SnowflakeSession.py
@@ -90,6 +90,7 @@ class SnowflakeSession:
             "private_key_file": kwargs.get(
                 "private_key_file"
             ),  # Already mapped from private_key_path in get_session_kwargs()
+            "private_key_file_pwd": kwargs.get("private_key_file_pwd"),
             "token": kwargs.get("token"),
             "password": kwargs.get("password"),
             "authenticator": kwargs.get("authenticator"),


mibelbahri (approved):
Good catch, thank you 🚀

sfc-gh-tmathew (approved):
Thank you for your contribution. 4.2.0 will include this contribution and additional tests to cover this scenario.

PR #411, opened 2026-02-09
Release 4.3.0

### LINKED ISSUE #214: Fail Upon Skipped Versioned Migration
# The Problem
I use the epoch time as the version number when creating versioned scripts. This _almost_ removes the version-number guesswork from the creation of a new script. Here's the exception:

Two developers are creating scripts around the same time. Or, one dev starts work on a script and has to come back to it after making some other small change:
- V1702512738__some_change_requiring_a_lengthy_review.sql
- V1702512750__a_small_change.sql

Because it is smaller or easier to review, the `V1702512750` script is applied by Schemachange first. When the `V1702512738` script is merged into the branch in question, Schemachange **silently** ignores it. I mean, it isn't silent. Somewhere in the thousands of lines of logging, there's a line about it. The dev _could_ check the logs or review Snowflake to ensure that the changes "took", but I think we can do better than that. 

# The Ask
I'd like to  support a flag that would cause schemachange to fail if both of the following are true:
- A versioned script isn't already applied **and**
- The script won't be applied because the max version number is already greater than the script version

In `fetch_change_history`, we're already fetching information from the `change_history_table` about versioned scripts. Instead of `get_alphanum_key(script['script_version']) <= get_alphanum_key(max_published_version)`, we could retrieve the `change_history_table` record when iterating over the repository scripts. If there's no change history table record to speak of and the "fail on missing" flag is true, throw an exception. While we're at it, we could compare the has of the versioned script to what was recorded in the change history table. If that value differs, a warning would suffice. 

If you're assigning arbitrary version numbers (e.g. 12345), this functionality is baked in. The process of provisioning a version number would likely involve reviewing _existing_ script version numbers and selecting the next available number. If two developers use this process and arrive at the same number, Schemachange will complain about the duplicated version number and fail the second migration. 

I can implement this in a fork, but I'd like to do so with feedback. Useful? Monstrous? Thanks for reading!

sfc-gh-tmathew:
## What does this PR do?

Release 4.3.0 with the following changes:

### Added
- **Connector Upgrade**: Bumped minimum snowflake-connector-python version from >=2.8 to >=3.0 (dropping Python connector 2.x support)
- **CLI Migration Scripts (.cli.yml)**: Execute CLI commands as part of your deployment process. Deploy complex Snowflake objects like dbt projects or Snowpark functions using the Snowflake CLI (snow) directly from schemachange
- **Out-of-Order Execution (--out-of-order)**: Allow versioned scripts to be applied regardless of whether their version is older than the maximum applied version
- **Failed Script Logging**: Script execution failures are now recorded in the change history table with STATUS = Failed

### Fixed
- **Password from connections.toml not being passed to Snowflake connector**
- Trailing comment handling regression from 4.2.0

## Checklist

- [x] Tests pass locally (pytest)
- [x] Code is formatted (ruff format .)
- [x] Ive added tests for new functionality
- [x] Ive updated relevant documentation

---

**Questions?** Just ask in the PR comments - were here to help!


sfc-gh-twhite (approved):
These changes look good, thank you!

At some point in the future we can likely remove the setup.cfg file to avoid maintenance in both the pyproject and that one.

TROUBLESHOOTING.md
```diff
@@ -364,12 +364,14 @@ export SNOWFLAKE_PASSWORD="my_password"
...
+### Error: `Script rendered to empty content` or `Script contains only comments or semicolons` (Issue #258)
 
-**Root Cause:** After Jinja template processing, the script contains only whitespace or comments that Snowflake's connector strips before execution.
+**Root Cause:** After Jinja template processing, the script contains only whitespace, comments, or semicolons.
 
-**How Schemachange Fixes This:**
-- ✅ **Valid SQL + trailing comments**: Auto-appends `SELECT 1;` no-op statement (metadata preserved, debug log message shown)
+**How Schemachange Handles This:**
+- ✅ **Valid SQL**: Passes through unchanged
+- ✅ **Trailing comments on new lines after `;`**: Auto-appends `SELECT 1;` to prevent empty statement error
```
sfc-gh-twhite, line 373: This is a very nice QoL change!

cc: @sfc-gh-sabbas 👀

PR #419, opened 2026-02-11
optimizing actions workflow

sfc-gh-tmathew:
This is more of an operational commit. Not related to functionailty or features of schemachange.


sfc-gh-swalia (approved):
Thanks, Tiji! This will help to reduce concurrent runs.

Issue #409, opened 2026-01-29
Case sensitivity of the config file


Dikootje:
**Describe the bug**
When running schemachange it deploys everyting everytime it runs. 
I get a lot of duplicate versions in the SCHEMACHANGE_CHANGE_HISTORY table. 

Turns out the change-history-table parameter is case sensitive. Snowflake by default stores everything in uppercase. 

**To Reproduce**
- I've created a `schemachange-config.yml` file with `change-history-table: "{{ env_var('SNOWFLAKE_DATABASE') }}.public.schemachange_change_history"`
- Run `schemachange deploy`, it should all work fine. 
- Check the SCHEMACHANGE_CHANGE_HISTORY table to verify that records are being logged. 
- Run `schemachange deploy` again, it should all work fine again, but notice how it is deploying all the files again. 
- Check the SCHEMACHANGE_CHANGE_HISTORY table to verify that there are duplicate records. 
- Change `schemachange-config.yml` file with to UPPERCASE `change-history-table: "{{ env_var('SNOWFLAKE_DATABASE') }}.PUBLIC.SCHEMACHANGE_CHANGE_HISTORY"`
- Run `schemachange deploy`, it should now skip previously deployed items. . 

**Expected behavior**
I expect the case not to matter in my configuration script as long there's no quotes around the object names. 

**Schemachange (please complete the following information):**
- Version (e.g. 3.5.3): 4.2.0

cburyta:
Ran into this recently too. I believe this existed for a while but it's popping up now per a change to get_script_metadata ([v4.1.0](https://github.com/Snowflake-Labs/schemachange/blob/v4.1.0/schemachange/session/SnowflakeSession.py#L241-L260), [v4.3.1](https://github.com/Snowflake-Labs/schemachange/blob/v4.3.1/schemachange/session/SnowflakeSession.py#L347-L385)).

Maybe make the [fetch_change_history_metadata](https://github.com/Snowflake-Labs/schemachange/blob/3b727425c08cf360461f7e9028e0ee94ee65b19d/schemachange/session/SnowflakeSession.py#L262) check case-insensitive?

PR #422, opened 2026-02-13
fix: avoid re-running run-once migrations (409)

cburyta:
# Pull Request: Fix INFORMATION_SCHEMA case sensitivity (#409)

## Summary

Fixes the bug where schemachange re-deploys all versioned scripts when the change history table is specified with lowercase schema/table names (e.g. `mydb.myschema.change_history`). Snowflake stores identifiers in **uppercase** in `INFORMATION_SCHEMA`; the previous query used a case-sensitive comparison, so the lookup returned no rows and schemachange treated the table as missing.

## Problem

- **Issue:** [#409 – Case sensitivity of the config file](https://github.com/Snowflake-Labs/schemachange/issues/409)
- **Symptom:** "When running schemachange it deploys everything every time. I get a lot of duplicate versions in the SCHEMACHANGE_CHANGE_HISTORY table."
- **Root cause:** `fetch_change_history_metadata()` and `change_history_schema_exists()` compared against `INFORMATION_SCHEMA` using `REPLACE('schema_name','"','')` without normalizing case. Snowflake returns `TABLE_SCHEMA = 'MYSCHEMA'`, so `'myschema' != 'MYSCHEMA'` and the query returns no rows.

## Solution

Use `UPPER(REPLACE(...))` in the WHERE clause for:

1. **`fetch_change_history_metadata()`** – `TABLE_SCHEMA` and `TABLE_NAME` in the `INFORMATION_SCHEMA.TABLES` query.
2. **`change_history_schema_exists()`** – `SCHEMA_NAME` in the `INFORMATION_SCHEMA.SCHEMATA` query.

No change to object references (e.g. `fully_qualified`); Snowflake is already case-insensitive for those. Only the INFORMATION_SCHEMA lookups are fixed.

## Testing

- **New test:** `test_fetch_change_history_metadata_uses_upper_for_information_schema` – builds a session with lowercase `ChangeHistoryTable` and asserts the executed query contains `UPPER(REPLACE(` so the lookup is case-insensitive.

## Checklist

- [x] Tests pass (`pytest`)
- [x] Code formatted (`ruff format .`)
- [x] Lint clean (`ruff check .`)
- [x] Test added for the fix
- [ ] CHANGELOG.md

### DIFF
--- a/schemachange/session/SnowflakeSession.py
+++ b/schemachange/session/SnowflakeSession.py
@@ -254,8 +254,7 @@ class SnowflakeSession:
         if self.change_history_table is None:
             raise ValueError("change_history_table is required for deployment operations")
         # This should only ever return 0 or 1 rows.
-        # Use UPPER() so lookup is case-insensitive: Snowflake stores identifiers in
-        # uppercase in INFORMATION_SCHEMA (fixes #409).
+        # UPPER() for case-insensitive match with INFORMATION_SCHEMA.
         query = f"""\
             SELECT
                 CREATED,
@@ -278,7 +277,7 @@ class SnowflakeSession:
     def change_history_schema_exists(self) -> bool:
         if self.change_history_table is None:
             raise ValueError("change_history_table is required for deployment operations")
-        # Use UPPER() for case-insensitive match with INFORMATION_SCHEMA (#409).
+        # UPPER() for case-insensitive match with INFORMATION_SCHEMA.
         query = f"""\
             SELECT
                 COUNT(1)
--- a/schemachange/session/SnowflakeSession.py
+++ b/schemachange/session/SnowflakeSession.py
@@ -253,14 +253,16 @@ class SnowflakeSession:
     def fetch_change_history_metadata(self) -> dict:
         if self.change_history_table is None:
             raise ValueError("change_history_table is required for deployment operations")
-        # This should only ever return 0 or 1 rows
+        # This should only ever return 0 or 1 rows.
+        # Use UPPER() so lookup is case-insensitive: Snowflake stores identifiers in
+        # uppercase in INFORMATION_SCHEMA (fixes #409).
         query = f"""\
             SELECT
                 CREATED,
                 LAST_ALTERED
             FROM {self.change_history_table.database_name}.INFORMATION_SCHEMA.TABLES
-            WHERE TABLE_SCHEMA = REPLACE('{self.change_history_table.schema_name}','\"','')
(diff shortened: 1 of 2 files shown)


sfc-gh-tmathew:
Thanks for the PR! One thing to flag on the `UPPER()` approach: it breaks quoted mixed-case identifiers. When someone specifies `"MySchema"`, `get_snowflake_identifier_string` stores it as `"MySchema"`, and `REPLACE('"MySchema"', '"', '')` correctly yields `MySchema`. Adding `UPPER()` turns that into `MYSCHEMA`, but `INFORMATION_SCHEMA` stores quoted identifiers in their original case — so the lookup fails.

Rather than changing the SQL, would you be open to fixing this in `get_snowflake_identifier_string` in `config/utils.py`? Unquoted identifiers that match the word pattern should be uppercased at parse time, since that's Snowflake's own normalization behavior:
```python
elif snowflake_identifier_pattern.match(input_value):
    return input_value.upper()
```
That handles both cases correctly without needing any changes to the queries.

cburyta:
Glad to go that route editing in the python side - that would feel cleaner.

However a couple notes...
- if `QUOTED_IDENTIFIERS_IGNORE_CASE=TRUE` is for some reason set in the Snowflake env, I don't believe that edge case is covered without some extra thought
- I believe hitting that method will impact more than just this migration. I might approach it different though, I think instead of `return input_value.upper()`

[DeployConfig.py](https://github.com/Snowflake-Labs/schemachange/blob/1df02ea36286f4c50671f38a52236638ab30895a/schemachange/config/DeployConfig.py#L68) might change the case of other props for example. (Edit: I might consider editing [ChangeHistoryTable.py from_str](https://github.com/Snowflake-Labs/schemachange/blob/1df02ea36286f4c50671f38a52236638ab30895a/schemachange/config/ChangeHistoryTable.py#L26) for a python based approach.)

sfc-gh-tmathew:
Thanks @cburyta for the contribution! Closing this PR as superseded by #440 (merged).

PR #440 fixed the same root cause — but at the right layer: `_normalize_identifier()` in `ChangeHistoryTable.from_str()` uppercases unquoted identifiers at parse time, which is where the fix belongs.

The `UPPER()` wrapping added in this PR would also introduce a regression: after `REPLACE('"mySchema"', '"', '')`, `UPPER()` converts `mySchema` → `MYSCHEMA` — but Snowflake stores quoted-identifier names in their original case in `INFORMATION_SCHEMA`. Users with quoted mixed-case change history table names would get silent lookup misses.

Thanks again for spotting the issue — the diagnosis in the PR description was spot on.

PR #339, opened 2025-08-02
feat: continue-on-error flags for repeatable and always scripts

luisggc:
## Summary

- Adds `--continue-repeatable-on-error` and `--continue-always-on-error` flags (and `--continue-all-on-error` as a shorthand for both)
- When set, schemachange continues executing remaining scripts of the respective type after a failure instead of stopping immediately
- Versioned scripts always stop on failure regardless of these flags
- Failed scripts are recorded in the change history table with full error messages and summarized at the end

## Configuration

| Parameter | Description |
|-----------|-------------|
| `--continue-all-on-error` | Continue executing remaining repeatable and always scripts even if one fails (shorthand for both flags below) |
| `--continue-repeatable-on-error` | Continue executing remaining repeatable scripts after an error |
| `--continue-always-on-error` | Continue executing remaining always scripts after an error |

Default behavior (fail-fast) is unchanged unless one of these flags is explicitly set.

## Notes

- Depends on #426 (change history table schema validation + `ERROR_MESSAGE` column) which should be merged first
- `--continue-all-on-error` expands to `continue_repeatable_on_error=True` + `continue_always_on_error=True` at config merge time
- The no-jinja marker (`-- schemachange-no-jinja`) has been moved to a separate PR

PR #426, opened 2026-03-26
feat: validate change history table schema upfront, add ERROR_MESSAGE column

luisggc:
## Summary

- Adds `ERROR_MESSAGE VARCHAR` to the `CREATE TABLE` statement so newly created change history tables include the column from the start
- Adds `validate_change_history_schema()` called at initialization (after the table is confirmed to exist, before any scripts run), using a fast metadata-only `SELECT ERROR_MESSAGE FROM <table> LIMIT 0` check
- If the column is missing on an existing table (created by an older version of schemachange or manually by a DBA): schemachange attempts `ALTER TABLE ... ADD COLUMN IF NOT EXISTS ERROR_MESSAGE VARCHAR` automatically
- If the role lacks ALTER privileges, the deploy fails immediately with a clear message showing the exact SQL to run manually — no mid-deploy surprises
- In dry-run mode, a warning is logged if the column is missing but no ALTER is attempted
- Removes the previous runtime try/except hack in `record_change_history` that added the column on INSERT failure

## Motivation

Previously, `ERROR_MESSAGE` was added via auto-ALTER when an INSERT failed at runtime. This was risky: it introduced DDL mid-deploy, and if the role lacked ALTER privileges the deploy would break even for successful scripts (since `error_message=''` is always passed). Moving this check to initialization gives a predictable, upfront failure instead.

This is a prerequisite for the continue-on-error feature (#339).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,10 +4,18 @@ All notable changes to this project will be documented in this file.
 *The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).*
 
 ## [Unreleased]
+### Added
+- **Error messages in change history**: Failed scripts now record the Snowflake error text in a new `ERROR_MESSAGE` column of the change history table (alongside the existing `STATUS`), making failures diagnosable directly from the change history.
+- **Upfront change history schema validation**: Before running any scripts, schemachange verifies the change history table has the required columns. If a table created by an older version is missing `ERROR_MESSAGE`, the column is added automatically on the first deploy (`ALTER TABLE ... ADD COLUMN IF NOT EXISTS ERROR_MESSAGE VARCHAR`). In `--dry-run` mode it only warns.
+
 ### Fixed
 - **R script checksum mismatch** ([#435](https://github.com/Snowflake-Labs/schemachange/issues/435)): Fixed a bug where R scripts with trailing comments (e.g. `-- comment` after the last `;`) were re-applied on every deployment even when unchanged. The root cause was `apply_change_script()` recomputing the checksum on post-transformation content instead of using the pre-transformation checksum from `deploy.py`.
 
 ### Upgrade Notes
+- **New `ALTER TABLE` privilege requirement:** The deploying role now needs `ALTER TABLE` on the change history table (previously only `SELECT`/`INSERT` were required), so schemachange can add the new `ERROR_MESSAGE` column on the first deploy after upgrading. Least-privilege deployments will fail on that first run unless the role has `ALTER TABLE`, or an operator adds the column manually beforehand:
+  ```sql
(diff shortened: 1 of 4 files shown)


sfc-gh-tmathew:
Thanks @luisggc — storing `ERROR_MESSAGE` in change history and validating the schema upfront are both genuinely useful. The feature concept and the `validate_change_history_schema` design are solid. A few things to address before this can merge:

## Blockers

**1. Rebase required**
This PR predates #437 (merged), which changed `apply_change_script()`. The diff no longer applies cleanly against current master. Please rebase.

**2. `ProgrammingError` catch is too broad**
`validate_change_history_schema` catches all `ProgrammingError` to detect a missing column, but Snowflake also raises `ProgrammingError` for a misconfigured schema name, wrong database, permission errors, etc. Those would be silently caught and then produce a confusing "please run ALTER manually" message instead of the real root cause. Fix: narrow the catch to error 904 (invalid identifier) or check `"invalid identifier" in str(e).lower()`.

**3. Migration docs — new privilege requirement**
This PR introduces a new requirement: the deploying role needs `ALTER TABLE` on the change history table (not just `INSERT`/`SELECT`). Existing least-privilege deployments will fail on the first run after upgrading. Needs a CHANGELOG entry and upgrade note explaining:
- The `ERROR_MESSAGE` column is auto-added on first deploy
- The deploying role must have `ALTER TABLE` privilege, or run this manually:
  `ALTER TABLE <change_history_table> ADD COLUMN IF NOT EXISTS ERROR_MESSAGE VARCHAR;`

**4. `get_script_metadata` tests need updating**
The PR inserts `self.validate_change_history_schema(dry_run=dry_run)` into `get_script_metadata`, but existing tests for that method don't mock `validate_change_history_schema`. Should add at least one test asserting it is called with the correct `dry_run` value.

## Nice to have

**5. `execution_time` now includes session-reset overhead**
Moving `start = time.time()` before the `if len(script_content) > 0` block and computing `execution_time` after `finally` means stored execution time now includes `reset_query_tag()` + `reset_session()` round-trips. Not wrong, but worth a comment in the code if intentional.

Once 1–4 are addressed, this is a good candidate for the next release cycle.

luisggc:
@sfc-gh-tmathew , Thanks for the review (again)! All four blockers and the nice-to-have are addressed:

1. **Rebase** — rebased onto current master; the `apply_change_script` checksum change and this feature coexist cleanly.
2. **ProgrammingError catch** — narrowed to error 904 / "invalid identifier" only. Wrong db/schema, permission, and other errors now re-raise with their real root cause instead of being masked.
3. **Migration docs** — added a CHANGELOG entry and upgrade note covering the new `ALTER TABLE` privilege requirement and the manual `ALTER TABLE ... ADD COLUMN IF NOT EXISTS ERROR_MESSAGE VARCHAR;` fallback; also updated the README privileges section and change-history DDL.
4. **Tests** — added a test asserting `get_script_metadata` calls `validate_change_history_schema` with the correct `dry_run`, plus one confirming a non-"invalid identifier" error propagates.
5. **execution_time** — added a comment noting it intentionally includes the session-reset overhead.

Full suite passes (611 tests), ruff clean. Ready for another look.

PR #427, opened 2026-03-26
feat: add schemachange-no-jinja marker to skip jinja rendering per script

luisggc:
## Summary

- Adds a `-- schemachange-no-jinja` marker: when present anywhere in a script file, schemachange skips jinja rendering entirely and executes the file as-is
- Useful for scripts that contain jinja-style syntax (e.g. `{{ }}`) that should be passed through to Snowflake rather than processed by schemachange
- No changes to behavior for scripts without the marker

### DIFF
--- a/README.md
+++ b/README.md
@@ -350,7 +350,7 @@ schemachange will replace any variable placeholders before running your change s
 finds any variable placeholders that haven't been replaced.
 
 If a script contains jinja-style syntax that should not be processed by schemachange, add the comment
-`--schemachange-no-jinja` anywhere in the file. When this marker is present, schemachange will skip jinja rendering for
+`-- schemachange-no-jinja` anywhere in the file. When this marker is present, schemachange will skip jinja rendering for
 that script and execute it as-is.
 
 #### Secrets filtering
--- a/README.md
+++ b/README.md
@@ -349,6 +349,10 @@ JSON object formatted as a string.
 schemachange will replace any variable placeholders before running your change script code and will throw an error if it
 finds any variable placeholders that haven't been replaced.
 
+If a script contains jinja-style syntax that should not be processed by schemachange, add the comment
+`-- schemachange-no-jinja` anywhere in the file. When this marker is present, schemachange will skip jinja rendering for
+that script and execute it as-is.
+
 #### Secrets filtering
 
 While many CI/CD tools already have the capability to filter secrets, it is best that any tool also does not output
--- a/README.md
+++ b/README.md
@@ -350,7 +350,7 @@ schemachange will replace any variable placeholders before running your change s
 finds any variable placeholders that haven't been replaced.
 
 If a script contains jinja-style syntax that should not be processed by schemachange, add the comment
-`-- schemachange-no-jinja` anywhere in the file. When this marker is present, schemachange will skip jinja rendering for
+`--schemachange-no-jinja` anywhere in the file. When this marker is present, schemachange will skip jinja rendering for
 that script and execute it as-is.
 
 #### Secrets filtering
--- a/README.md
+++ b/README.md
@@ -349,9 +349,10 @@ JSON object formatted as a string.
 schemachange will replace any variable placeholders before running your change script code and will throw an error if it
 finds any variable placeholders that haven't been replaced.
 
-If a script contains jinja-style syntax that should not be processed by schemachange, add the comment
-`-- schemachange-no-jinja` anywhere in the file. When this marker is present, schemachange will skip jinja rendering for
-that script and execute it as-is.
+If a script contains jinja-style syntax that should not be processed by schemachange, add the marker
+`schemachange-no-jinja` in a SQL line comment (e.g. `-- schemachange-no-jinja`). When this marker is present,
+schemachange skips jinja rendering for that script and executes it as-is. The marker is case-insensitive and only
+recognized in a `--` line comment, so the token inside a string literal or block comment does not disable jinja.
 
 #### Secrets filtering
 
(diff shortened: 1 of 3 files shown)


sfc-gh-tmathew:
Thanks @luisggc — great idea, `-- schemachange-no-jinja` fills a real gap for scripts that use `{{ }}` natively. Two issues to fix before this is ready:

## Required

**1. False-positive substring match**
`_NO_JINJA_MARKER in source.lower()` matches anywhere in the file — including inside SQL string literals or comments. A script like:
```sql
UPDATE config SET value = 'schemachange-no-jinja' WHERE key = 'feature_flag';
```
...would silently bypass Jinja for the entire file. Fix: restrict to line comments with a regex:
```python
import re
_NO_JINJA_PATTERN = re.compile(r'--[^\n]*schemachange-no-jinja', re.IGNORECASE)
```
Then use `_NO_JINJA_PATTERN.search(source)` instead of the bare `in` test.

**2. Double file read in the non-marker path (affects every script)**
`get_source()` reads the file unconditionally, then `get_template()` calls `get_source()` again internally via the loader. Every script that does not carry the marker now incurs two reads. Fix by using the already-read source to compile:
```python
source, _, _ = self.__environment.loader.get_source(self.__environment, posix_path)
if _NO_JINJA_PATTERN.search(source):
    raw_content = source
else:
    template = self.__environment.from_string(source)
    raw_content = template.render(**variables)
```
Note: `from_string()` produces an anonymous template so Jinja error messages won't include the filename — acceptable trade-off, or set `name=posix_path` explicitly.

Once those two are addressed this is clean for inclusion in an upcoming release.

luisggc:
Thanks for the review, @sfc-gh-tmathew. Both issues are addressed in the latest push.

**1. False-positive substring match**

Switched to a comment-scoped, case-insensitive regex so the marker is only honored inside a `--` line comment:

```python
_NO_JINJA_PATTERN = re.compile(r"--[^\n]*schemachange-no-jinja", re.IGNORECASE)
...
if _NO_JINJA_PATTERN.search(source):
```

Your `UPDATE config SET value = 'schemachange-no-jinja' ...` example now renders normally — I added a test for exactly that (marker in a string literal → Jinja still runs), plus a case-insensitivity test.

**2. Double file read**

Now reading the source once and rendering it directly instead of calling `get_source()` and then `get_template()`:

```python
source, _, _ = self.__environment.loader.get_source(self.__environment, posix_path)
if _NO_JINJA_PATTERN.search(source):
    raw_content = source
else:
    template = self.__environment.from_string(source)
    raw_content = template.render(**variables)
```

Un-marked scripts now do a single read. On the filename trade-off you noted: I went with the plain `from_string()` (anonymous template) for now, since schemachange already logs the script path around render() — happy to switch to `env.compile(..., name=posix_path)` to keep the filename in Jinja tracebacks if you'd prefer.

Also updated the README to state the line-comment requirement (it previously said the marker worked "anywhere in the file").

atg-simon:
Hi, @luisggc @sfc-gh-tmathew would marking a block as raw using the native jinja functionality not get the desired effect. https://jinja.palletsprojects.com/en/stable/templates/#escaping

``` sql
{% raw %}
--SQL here is not processed by jinja
SELECT '{{'
{% endraw %}
```

sfc-gh-swalia:
> Hi, @luisggc @sfc-gh-tmathew would marking a block as raw using the native jinja functionality not get the desired effect. https://jinja.palletsprojects.com/en/stable/templates/#escaping
> 
> ```sql
> {% raw %}
> --SQL here is not processed by jinja
> SELECT '{{'
> {% endraw %}
> ```


Thanks for the work here @luisggc and thanks for your inputs @atg-simon. The implementation is clean after @sfc-gh-tmathew's feedback was addressed.

One thing I'd like us to discuss before merging:

**Overlap with `{% raw %}`**

`{% raw %}...{% endraw %}` already handles both the "whole file" case (wrap everything) and the "partial" case (wrap just the JS/non-Jinja block). The marker only handles whole files. This means users who start with `-- schemachange-no-jinja` will eventually hit a mixed scenario (e.g., need `{{ env }}` resolved but also have JavaScript `{{ }}`) and discover the marker doesn't help — they have to learn `{% raw %}` anyway.

I'm not opposed to merging this - it's a cleaner Developer experience for pure-SQL/JS files that never touch Jinja, and a SQL comment is friendlier to linters and editors than `{% raw %}`. But I want to make sure we're deliberate about expanding the API surface for a convenience feature.

**If we do merge, the README should document the limitation:**

> **Note:** `-- schemachange-no-jinja` disables Jinja for the *entire* file. If your script needs both Jinja variable substitution (e.g., `{{ env }}`) and literal curly braces (e.g., in JavaScript UDFs), use Jinja's built-in `{% raw %}...{% endraw %}` block around the literal portions instead:
>
> ```sql
> CREATE OR REPLACE FUNCTION {{ env }}.my_func()
> RETURNS STRING
> LANGUAGE JAVASCRIPT
> AS $$
> {% raw %}
>   return `{"time": "{{EVENT_TIME}}"}`;
> {% endraw %}
> $$;
> ```

Would appreciate thoughts from @sfc-gh-tmathew on whether this is worth the added surface area or if we'd rather point users to `{% raw %}` as the single documented solution.

README.md
```diff
@@ -356,11 +349,16 @@ JSON object formatted as a string.
 schemachange will replace any variable placeholders before running your change script code and will throw an error if it
 finds any variable placeholders that haven't been replaced.
 
+If a script contains jinja-style syntax that should not be processed by schemachange, add the marker
```
sfc-gh-swalia, line 352: Please refer to my comment [here](https://github.com/Snowflake-Labs/schemachange/pull/427/#issuecomment-5038702954)

PR #437, opened 2026-05-22
Fix: prevent R scripts with trailing comments from being reapplied (#…

### LINKED ISSUE #435: R script getting reapplied when there are trailing comments at the EOF
**Describe the bug**
All R scripts that contain trailing comments get redeployed every time my CI/CD pipeline runs. I'm sure that the files didn't change, and the checksum that shows in the change history table in Snowflake also didn't change for the affected files.

In my case, the R__ scripts are just defining tasks with `CREATE OR REPLACE TASK` statement.

**To Reproduce**
1. Create an R__ script with any statement and add a comment at the end of the file like "-- this is a dummy comment;" (not sure if the ; at the end is necessary, but it had in my case)
2. Create a PR and merge it
3. Run the CI/CD pipeline to deploy the file
4. Merge another PR with any other change, like add a V script or another R script (do not change anything on the one that has the trailing comment)
5. Run the CI/CD pipeline and see the R__ script getting redeployed

**Expected behavior**
The R script not getting reapplied if it doesn't receive any changes

**Screenshots**

Image

**Schemachange (please complete the following information):**
- Version 4.3.2

**Additional context**
I'm using an Azure DevOps CI/CD pipeline for the deployment.
Snowflake Edition: Business Critical
Cloud platform: Azure

sfc-gh-swalia:
## What does this PR do?

Fixes #435 — R scripts with trailing comments after the last semicolon get reapplied on every deployment even when unchanged.

Closes #(435)
## Root Cause

`prepare_for_execution()` appends `SELECT 1; -- schemachange: trailing comment fix` to prevent Snowflake's "Empty SQL Statement" error. However, `apply_change_script()` in `SnowflakeSession.py` was **recomputing** the checksum on this modified content before storing it in the change history table.

On subsequent runs, `deploy.py` computes the checksum on the **original** rendered content (pre-transformation) and compares it against the stored post-transformation checksum — they never match, causing infinite reapplication.

The CLI script path (`deploy.py:152`) was already correct — it passes `checksum_current` directly to `record_change_history()`.

## Fix

- Added optional `checksum: str | None = None` parameter to `apply_change_script()`
- `deploy.py` now passes the pre-computed `checksum_current` to `apply_change_script()`
- If `checksum` is not provided (backward compat), it falls back to computing it from `script_content`

## For Testing: 
- Reproduced the bug with an R script containing trailing comments (v4.3.2)
- Verified the CLI `.cli.yml` path does NOT have this bug (correct baseline)
- Applied fix → second deploy now correctly skips unchanged R scripts
- All 602 unit tests pass

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,13 @@ All notable changes to this project will be documented in this file.
 
 *The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).*
 
+## [Unreleased]
+### Fixed
+- **R script checksum mismatch** ([#435](https://github.com/Snowflake-Labs/schemachange/issues/435)): Fixed a bug where R scripts with trailing comments (e.g. `-- comment` after the last `;`) were re-applied on every deployment even when unchanged. The root cause was `apply_change_script()` recomputing the checksum on post-transformation content instead of using the pre-transformation checksum from `deploy.py`.
+
+### Upgrade Notes
+- **Note for users with R scripts that have trailing comments:** If you deployed R scripts with trailing comments under v4.3.x, their checksums stored in the change history table were computed from the post-transformation content. On the **first deploy after upgrading**, those scripts will be re-applied once (schemachange will see a checksum mismatch). For idempotent scripts this is harmless; non-idempotent R scripts should be reviewed before upgrading.
+
 ## [4.3.3] - 2026-04-20
 ### Fixed
 - Relaxed structlog version constraint to `>=24.1.0,<26.0` to support structlog 25.x ([#429](https://github.com/Snowflake-Labs/schemachange/pull/429) by [@Falydoor](https://github.com/Falydoor)).
--- a/schemachange/deploy.py
+++ b/schemachange/deploy.py
@@ -174,6 +174,7 @@ def deploy(config: DeployConfig, session: SnowflakeSession):
                 dry_run=config.dry_run,
                 logger=script_log,
                 out_of_order=is_out_of_order,
+                checksum=checksum_current,
             )
 
         scripts_applied += 1
--- a/schemachange/session/SnowflakeSession.py
+++ b/schemachange/session/SnowflakeSession.py
@@ -472,6 +472,7 @@ class SnowflakeSession:
         dry_run: bool,
         logger: structlog.BoundLogger,
         out_of_order: bool = False,
+        checksum: str | None = None,
     ) -> None:
         if self.change_history_table is None:
             raise ValueError("change_history_table is required for deployment operations")
@@ -485,7 +486,8 @@ class SnowflakeSession:
             logger.info(f"Applying {script.type_desc} change script")
         # Define a few other change related variables
         # noinspection PyTypeChecker
-        checksum = hashlib.sha224(script_content.encode("utf-8")).hexdigest()
+        if checksum is None:
+            checksum = hashlib.sha224(script_content.encode("utf-8")).hexdigest()
         execution_time = 0
 
         # Execute the contents of the script
--- a/tests/session/test_SnowflakeSession.py
+++ b/tests/session/test_SnowflakeSession.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import hashlib
 from collections import defaultdict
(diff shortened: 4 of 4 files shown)


sfc-gh-tmathew:
Thanks @sfc-gh-swalia — the root cause analysis and fix are correct. We'd like to include this in rc4.4.0, but have two asks before we merge:

## Required: Regression test for the checksum path

The fix is a 4-line change and the bug is subtle — exactly the kind of thing that could regress silently. Please add a unit test to cover it. Minimum suggested:

```python
def test_apply_change_script_uses_provided_checksum(mock_session):
    """apply_change_script should store the caller-supplied checksum, not recompute from script_content."""
    pre_transform_checksum = hashlib.sha224(b"SELECT 1;\n-- trailing comment").hexdigest()
    post_transform_content = "SELECT 1;\n-- trailing comment\nSELECT 1; -- schemachange: trailing comment fix"

    with mock.patch.object(mock_session, "record_change_history") as mock_record:
        mock_session.apply_change_script(
            script=...,
            script_content=post_transform_content,
            checksum=pre_transform_checksum,
            dry_run=False,
            logger=...,
        )
    assert mock_record.call_args.kwargs["checksum"] == pre_transform_checksum
```

A second test covering the R script deploy loop would be even better (deploy twice, verify second run skips).

## Should have: CHANGELOG note

Please add an entry under the next section in CHANGELOG.md calling out the one-time migration impact:

> **Note for users with R scripts that have trailing comments:** If you deployed R scripts with trailing comments (e.g. `-- comment` after the last `;`) under v4.3.x, their checksums stored in the change history table were computed from the post-transformation content. On the **first deploy after upgrading**, those scripts will be re-applied once (schemachange will see a checksum mismatch). For idempotent scripts this is harmless; non-idempotent R scripts should be reviewed before upgrading.

Once those two items are addressed, we'll fast-track this into rc4.4.0.

sfc-gh-swalia:
> Thanks @sfc-gh-swalia — the root cause analysis and fix are correct. We'd like to include this in rc4.4.0, but have two asks before we merge:
> 
> ## Required: Regression test for the checksum path
> The fix is a 4-line change and the bug is subtle — exactly the kind of thing that could regress silently. Please add a unit test to cover it. Minimum suggested:
> 
> ```python
> def test_apply_change_script_uses_provided_checksum(mock_session):
>     """apply_change_script should store the caller-supplied checksum, not recompute from script_content."""
>     pre_transform_checksum = hashlib.sha224(b"SELECT 1;\n-- trailing comment").hexdigest()
>     post_transform_content = "SELECT 1;\n-- trailing comment\nSELECT 1; -- schemachange: trailing comment fix"
> 
>     with mock.patch.object(mock_session, "record_change_history") as mock_record:
>         mock_session.apply_change_script(
>             script=...,
>             script_content=post_transform_content,
>             checksum=pre_transform_checksum,
>             dry_run=False,
>             logger=...,
>         )
>     assert mock_record.call_args.kwargs["checksum"] == pre_transform_checksum
> ```
> 
> A second test covering the R script deploy loop would be even better (deploy twice, verify second run skips).
> 
> ## Should have: CHANGELOG note
> Please add an entry under the next section in CHANGELOG.md calling out the one-time migration impact:
> 
> > **Note for users with R scripts that have trailing comments:** If you deployed R scripts with trailing comments (e.g. `-- comment` after the last `;`) under v4.3.x, their checksums stored in the change history table were computed from the post-transformation content. On the **first deploy after upgrading**, those scripts will be re-applied once (schemachange will see a checksum mismatch). For idempotent scripts this is harmless; non-idempotent R scripts should be reviewed before upgrading.
> 
> Once those two items are addressed, we'll fast-track this into rc4.4.0.

I have addressed your comment. Thanks for taking the time to review.

sfc-gh-tmathew (approved):
Looks great @sfc-gh-swalia — both asks addressed cleanly. Tests cover the checksum passthrough and the fallback path. CHANGELOG note clearly documents the migration impact. Merging into rc4.4.0.

txdataguy:
@sfc-gh-tmathew we appear to be encountering this issue in schemachange 4.3.3, where R__ scripts with trailing comments are reapplied even when the checksum has not changed. It was addressed here with PR https://github.com/Snowflake-Labs/schemachange/pull/437. I assume next release will include this fix. Do we know when 4.4.0 will be available?

PR #439, opened 2026-05-22
Improve Quick Start first-deploy experience and add Staying Current Section

sfc-gh-swalia:
Quick Start: Add --create-change-history-table to all deploy examples and
a note explaining that the target database must already exist.

Staying Current: New section advising users to pin versions in CI, use
Dependabot/Renovate, and read the CHANGELOG before upgrading.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,10 @@ EOF
 
 ### 3. Run your first deployment
 
+> **First time?** The examples below include `--create-change-history-table`, which creates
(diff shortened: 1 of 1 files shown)


sfc-gh-tmathew (approved):
LGTM — adds --create-change-history-table to all Quick Start deploy examples and a clear first-deploy note. The Staying Current section is a good addition for CI/CD users.

PR #440, opened 2026-05-29
fix: normalize unquoted identifiers to uppercase

### LINKED ISSUE #432: R__ scripts re-applied on every deploy when change-history-table uses lowercase identifiers (case-sensitive INFORMATION_SCHEMA lookup)
**Describe the bug**

When `change-history-table` is configured with lowercase schema/table identifiers (e.g. `MY_DB.tech.schemachange_logs`), schemachange fails to recognize the existing history table for the purposes of skip logic, and **every `R__` (Repeatable) script is re-applied on every deploy** even when the rendered checksum is unchanged.

Write path still works (Snowflake resolves the identifier case-insensitively), so the history table keeps filling with duplicate rows that all carry the same checksum. The skip-by-checksum feature introduced in 3.0.0 (#49) effectively breaks.

Root cause appears to be the pre-deploy existence probe, which passes the schema and table names straight from the config into a `WHERE` clause as string literals against `INFORMATION_SCHEMA.TABLES`. Snowflake stores unquoted identifiers uppercase in metadata, and string comparison in the predicate is case-sensitive, so the lookup returns 0 rows when the config is lowercase:

```sql
SELECT CREATED, LAST_ALTERED
FROM MY_DB.INFORMATION_SCHEMA.TABLES
WHERE TABLE_SCHEMA = REPLACE('tech','"','')
  AND TABLE_NAME   = REPLACE('schemachange_logs','"','')
```

Downstream lookup of prior checksums per script then behaves as if no history existed, so every `R__` is treated as new.

**To Reproduce**

1. Create a schemachange project with a `R__` script and a config whose `change-history-table` uses lowercase identifiers:
   ```yaml
   config-version: 1
   change-history-table: "MY_DB.tech.schemachange_logs"
   create-change-history-table: true
   ```
2. `schemachange deploy` — first run applies the R__ script. Table gets auto-created as `MY_DB.TECH.SCHEMACHANGE_LOGS` in Snowflake.
3. Do not change anything.
4. `schemachange deploy` a second time.
5. Observe the `R__` script is re-applied (`scripts_applied=1 scripts_skipped=0`) with a new row in `change_history` carrying the **same** `CHECKSUM`.
6. Change `change-history-table` to uppercase (`MY_DB.TECH.SCHEMACHANGE_LOGS`). Re-run deploy — script is correctly skipped (`scripts_applied=0 scripts_skipped=1`).

**Expected behavior**

Skip behavior should not depend on the case of identifiers in `change-history-table`. Snowflake itself treats unquoted identifiers as case-insensitive, and the README makes no mention of a case requirement.

**Schemachange (please complete the following information):**
- Version: **4.3.3**
- Python: 3.11 / 3.12
- Snowflake edition: Enterprise

**Additional context**

Verbose log from a deploy that wrongly re-applies (lowercase config):

```
[info] Using existing change history table MY_DB.TECH.SCHEMACHANGE_LOGS
[debug] Executing query
  SELECT CREATED, LAST_ALTERED
  FROM MY_DB.INFORMATION_SCHEMA.TABLES
  WHERE TABLE_SCHEMA = REPLACE('tech','"','')
    AND TABLE_NAME   = REPLACE('schemachange_logs','"','')
...
[info] Applying R SQL change script   a_script_name=R__0000_cicd_role.sql script_version=N/A
[info] Completed successfully         scripts_applied=7 scripts_skipped=0
```

vs. identical repo with the only difference being uppercase identifiers in `change-history-table`:

```
[debug] Skipping change script because there is no change since the last execution
[info] Completed successfully         scripts_applied=0 scripts_skipped=7
```

Rendered script checksum (computed by `schemachange render`) equals the `CHECKSUM` already stored in `change_history` in both cases:
```
0dcd86705eaed581d1d7abd80395927a6526bb8ac29722736be00ab0
```

**Suggested fix**

Either uppercase the identifier components before the `INFORMATION_SCHEMA` predicate, or use `UPPER(TABLE_SCHEMA) = UPPER(...)`. Equivalent treatment may be needed everywhere identifiers from config feed into metadata-backed string comparisons.

sfc-gh-swalia:
## What does this PR do?

Unquoted identifiers in change-history-table config (e.g. mydb.myschema.mytable) were compared case-sensitively against INFORMATION_SCHEMA, which stores names in uppercase. This caused R__ scripts to be re-applied on every deploy when the config used lowercase.

Added ```_normalize_identifier()``` to ChangeHistoryTable.py that uppercases unquoted identifiers while preserving quoted ones. Updated the tests to reflect correct Snowflake identifier semantics.

Closes #432

## Checklist

### DIFF
--- a/schemachange/config/ChangeHistoryTable.py
+++ b/schemachange/config/ChangeHistoryTable.py
@@ -4,6 +4,19 @@ from typing import ClassVar
 from schemachange.config.utils import get_snowflake_identifier_string
 
 
+def _normalize_identifier(value: str) -> str:
+    """Uppercase unquoted identifiers to match Snowflake's INFORMATION_SCHEMA storage.
+
+    Snowflake stores unquoted identifiers in uppercase in metadata catalogs.
+    Quoted identifiers (e.g. "mySchema") are case-sensitive and left as-is.
+
+    Fixes: https://github.com/Snowflake-Labs/schemachange/issues/432
+    """
+    if value and not (value.startswith('"') and value.endswith('"')):
+        return value.upper()
+    return value
+
+
 @dataclasses.dataclass(frozen=True)
 class ChangeHistoryTable:
     _default_database_name: ClassVar[str] = "METADATA"
@@ -43,7 +56,13 @@ class ChangeHistoryTable:
                 raise ValueError(f"Invalid change history table name: {table_str}")
 
         return cls(
-            table_name=get_snowflake_identifier_string(input_value=table_name, input_type="table_name"),
-            schema_name=get_snowflake_identifier_string(input_value=schema_name, input_type="schema_name"),
-            database_name=get_snowflake_identifier_string(input_value=database_name, input_type="database_name"),
+            table_name=_normalize_identifier(
+                get_snowflake_identifier_string(input_value=table_name, input_type="table_name")
+            ),
+            schema_name=_normalize_identifier(
+                get_snowflake_identifier_string(input_value=schema_name, input_type="schema_name")
+            ),
+            database_name=_normalize_identifier(
+                get_snowflake_identifier_string(input_value=database_name, input_type="database_name")
+            ),
         )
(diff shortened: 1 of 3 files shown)


sfc-gh-tmathew (approved):
LGTM — fix is correct. Identifier normalisation order is provably right (get_snowflake_identifier_string first, then uppercase). No breaking change risk. Minor: consider updating the type annotation on _normalize_identifier to str | None -> str | None, but not a blocker.

PR #442, opened 2026-06-26
Removed --snowflake-password from demo example

sfc-gh-swalia:
## What does this PR do?

Removed ```--snowflake-password``` from demo CLI examples

Based on my analysis on commit history it seems that --snowflake-password was never implemented as a CLI argument (passwords are intentionally environemntal var only to avoid exposure in process lists and shell history). Updated the demo examples to use SNOWFLAKE_PASSWORD instead

## Checklist

### DIFF
--- a/demo/README.MD
+++ b/demo/README.MD
@@ -62,12 +62,16 @@ schemachange deploy --config-folder ./demo/basics_demo
 
 **Alternative: Using CLI arguments (NEW in 4.1.0)**
 ```bash
+# Set password via environment variable
+# schemachange will read this automatically during deploy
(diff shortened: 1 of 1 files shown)


sfc-gh-tmathew (approved):
LGTM. Both claims verified against live CLI:
- `--snowflake-password` does not exist as a CLI arg (confirmed via `uv run schemachange deploy --help`)
- Correct flag is `--schemachange-vars` / `--vars` (not `--schemachange-config-vars`)

Demo docs are now consistent with the actual implementation. Good catch.

PR #448, opened 2026-08-03
fix(security): escape single quotes in reset_query_tag (CWE-89) 

sfc-gh-swalia:
## What does this PR do?

**escape single quotes in reset_query_tag (CWE-89)**

```ALTER SESSION SET QUERY_TAG``` interpolated query_tag (from user config or
extra_tag from filenames) directly into SQL without escaping. A value
containing a single quote could break out of the string literal and
inject arbitrary SQL.

Apply standard SQL single-quote escaping (' -> '') before interpolation.

#446

## Checklist

### DIFF
--- a/schemachange/session/SnowflakeSession.py
+++ b/schemachange/session/SnowflakeSession.py
@@ -466,7 +466,9 @@ class SnowflakeSession:
         if extra_tag:
             query_tag += f";{extra_tag}"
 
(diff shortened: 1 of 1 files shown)


sfc-gh-tmathew (approved):
## Approved for 4.4.0

Simple, correct fix for CWE-89 finding from the security review (ref #446). Single-quote escaping is the appropriate mitigation for DDL statements like `ALTER SESSION SET` which don't support bind parameters.

Low risk, minimal change, addresses the flagged vulnerability. Merge to master for inclusion in 4.4.0.

PR #449, opened 2026-08-03
fix(security): use parameterized queries in record_change_history 

sfc-gh-swalia:
## What does this PR do? 

#446

The INSERT statement in record_change_history previously used f-string interpolation to embed script metadata (name, description, version) directly into SQL. Because execute_snowflake_query splits on semicolons, a crafted filename containing ```'; DROP TABLE ...' ``` could inject arbitrary SQL.

Replace f-string value interpolation with bind parameters (%s) passed to cursor.execute(), which prevents any value content from being interpreted as SQL syntax or statement separators.

## Checklist

### DIFF
--- a/schemachange/session/SnowflakeSession.py
+++ b/schemachange/session/SnowflakeSession.py
@@ -651,16 +651,26 @@ class SnowflakeSession:
                 STATUS,
                 INSTALLED_BY,
                 INSTALLED_ON
-            ) VALUES (
-                '{getattr(script, "version", "")}',
-                '{script.description}',
-                '{script.name}',
-                '{script.type}',
-                '{checksum}',
-                {execution_time},
-                '{status}',
-                '{self.user}',
-                CURRENT_TIMESTAMP
-            );
+            ) VALUES (%s, %s, %s, %s, %s, %s, %s, %s, CURRENT_TIMESTAMP)
         """
-        self.execute_snowflake_query(dedent(query), logger=logger)
+        params = (
+            getattr(script, "version", ""),
+            script.description,
+            script.name,
+            script.type,
+            checksum,
+            execution_time,
+            status,
+            self.user,
+        )
+        logger.debug("Recording change history", params=params)
+        try:
+            with self.con.cursor() as cur:
+                cur.execute(dedent(query), params)
+            if not self.autocommit:
+                self.con.commit()
+        except Exception as e:
+            logger.error("Failed to record change history", error_msg=str(e))
+            if not self.autocommit:
+                self.con.rollback()
+            raise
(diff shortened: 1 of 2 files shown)


sfc-gh-tmathew (changes requested):
## Request Changes — targeted for 4.4.0 with modifications

The parameterized query approach is the correct fix for CWE-89 in `record_change_history` (ref #446 security review). However, the implementation introduces a code path divergence that needs addressing before merge.

### Issue: bypasses session's execution pattern

The current implementation introduces direct cursor management (`with self.con.cursor()`) with its own commit/rollback logic, while every other method in the session uses `execute_snowflake_query` → `con.execute_string()`. This creates:

1. **Inconsistent error handling** — catches generic `Exception` vs `ProgrammingError` in the existing helper
2. **Separate commit/rollback logic** — duplicates what `execute_snowflake_query` already does
3. **Maintenance risk** — future session-level changes (logging, retries, connection pooling) won't automatically apply to this method

### Suggested approach

Since `execute_snowflake_query` doesn't support bind params (it uses `execute_string`), consider one of:

**Option A (preferred):** Add a lightweight helper method to the session class for parameterized single-statement execution:
```python
def _execute_parameterized(self, query: str, params: tuple, logger: structlog.BoundLogger):
    """Execute a single parameterized query with bind variables."""
    logger.debug("Executing parameterized query", query_preview=query[:200])
    try:
        with self.con.cursor() as cur:
            cur.execute(query, params)
        if not self.autocommit:
            self.con.commit()
    except snowflake.connector.errors.ProgrammingError as e:
        logger.error("SQL execution error", error_code=getattr(e, "errno", None), error_msg=str(e))
        if not self.autocommit:
            self.con.rollback()
        raise
```
Then `record_change_history` calls `self._execute_parameterized(query, params, logger)`.

**Option B:** Use the existing pattern but with `cursor.execute` directly in `record_change_history`, matching the existing error handling contract (`ProgrammingError`, not bare `Exception`).

### Also

- The `#446` issue reference is a security-review smoke test, not a feature issue. Consider creating a dedicated security issue for tracking this fix, or just document it in the commit message clearly.

### Tests

Test coverage is good — especially the injection regression test. Those should stay as-is regardless of the execution approach chosen.

---

This is targeted for 4.4.0 given the security-team finding. The requested changes are small — mainly wrapping the cursor logic to match existing patterns.

sfc-gh-tmathew (approved):
## Approved for 4.4.0

Requested changes addressed cleanly:

- `_execute_parameterized` helper now matches the session's existing error handling contract (`ProgrammingError` + `DatabaseError`, consistent commit/rollback)
- `record_change_history` delegates to the helper — no more code path divergence
- Tests remain comprehensive (injection regression test is solid)

Ready to merge to master for inclusion in rc4.4.0.

